PR #57, opened 2019-11-25
Add SwiftSyntax and skeleton code

elsh:
-Added --use-sourcekit as a command arg to toggle between sourcekit and swiftsyntax parser 
-Added SwiftSyntax framework dependency in package file 
-Added SwiftSyntaxExtension file with a placeholder Visitor
-Modified generator funcs with parser type input


Sources/Mockolo/Executor.swift
```diff
@@ -105,9 +105,9 @@ class Executor {
                                       shortName: "-j",
                                       kind: Int.self,
                                       usage: "Maximum number of threads to execute concurrently (default = number of cores on the running machine).")
-        version = parser.add(option: "--version",
-                             kind: String.self,
-                             usage: "Xcode version.")
+        parserType = parser.add(option: "--parser",
+                             kind: Int.self,
+                             usage: "Type of source file parser: 0 for SwiftSyntax, 1 for SourceKit (default = 0).")
```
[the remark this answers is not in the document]
elsh: I could but it's only between these two choices, and a user having to type in sourcekit / swiftsyntax also doesn't seem that user-friendly.

elsh:
> Too much code for me to review in one shot, but seems fine from an initial glance. Some questions:
> 
> * why keep supporting SourceKitten?

Keeping it for now in case we need to fall back to sourcekit (which shouldn't happen but just to be safe) and also makes it easier to measure performance difference just by toggling a parser type.


> * my reading of this is that the parsing is now single threaded, or is the concurrency happening somewhere I did not see?

It'll be added later. For now single threaded as SyntaxVisitor has mutating funcs and need to put a lock around them.

elsh:
> Might want to skip straight to 11.2, we're updating next week

I thought this is the minimum xcode version required, in which case it should probably be set to 11.0...

rmaz:
> > Might want to skip straight to 11.2, we're updating next week
> 
> I thought this is the minimum xcode version required, in which case it should probably be set to 11.0...

`osx_image` in the travis file configures what OS version is used to build with on Travis. 

```
platforms: [
        .macOS(.v10_14),
    ],	    
```

in `Package.swift` is what controls your deployment target.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -0,0 +1,352 @@
...
+            }
+            return AnnotationMetadata(typealiases: aliasMap)
+        }
+        return nil
+    }
+    func annotationMetadata(with annotation: String) -> AnnotationMetadata? {
+        var ret: AnnotationMetadata?
+        for i in 0..<count {
+            let trivia = self[i]
+            switch trivia {
```
mkruber: given similar functionality for `.docLineComment` and `.docBlockComment`, maybe this can be an `if` condition instead of a `switch`.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -0,0 +1,352 @@
...
+            }
+        }
+        return nil
+    }
+}
+
+extension Syntax {
+    var name: String {
+        if let decl = self as? ProtocolDeclSyntax {
+            return decl.identifier.description.trimmingCharacters(in: .whitespaces)
```
mkruber: Can the `token.text` not be used here?

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -0,0 +1,352 @@
...
+        return .visitChildren
+    }
+    
+    func visit(_ node: AttributeListSyntax) -> SyntaxVisitorContinueKind {
+        current?.attributes.append(node.withoutTrivia().description.trimmingCharacters(in: .whitespacesAndNewlines))
+        return .visitChildren
+    }
+    
+    func visit(_ node: VariableDeclSyntax) -> SyntaxVisitorContinueKind {
+        for v in node.bindings {
```
mkruber: some comments on what is being done here may be helpful.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -0,0 +1,352 @@
...
+                                        modelDescription: node.description,
+                                        processed: current?.isProcessed ?? false) {
+            current?.members.append(initmodel)
+        }
+        return .visitChildren
+    }
+    
+    func visit(_ node: FunctionDeclSyntax) -> SyntaxVisitorContinueKind {
+        var isStatic = false
+        if let mds = node.modifiers {
```
mkruber: this `if` condition can be refactored into a separate function and used for this visit as well as for visiting `VarDeclSyntax`.

Sources/Mockolo/Executor.swift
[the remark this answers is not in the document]
benpious: Most people won't run this manually though right? Seems like we should prioritize readability over convenience.

Sources/MockoloFramework/Operations/ProtocolMapGenerator.swift
```diff
@@ -16,94 +16,134 @@
...
+    case .sourceKit:
+        guard let annotationData = annotation.data(using: .utf8) else {
+            fatalError("Annotation is invalid: \(annotation)")
+        }
+        
+        if let queue = queue {
+            let lock = NSLock()
+            for filePath in files {
+                _ = semaphore?.wait(timeout: DispatchTime.distantFuture)
+                queue.async {
```
benpious: Seems like batch async processing is a common pattern in Mockolo, there should be a utility function for this similar to dispatch_function in GCD.

Sources/Mockolo/Executor.swift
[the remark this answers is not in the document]
elsh: It's a boolean toggle arg now

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
elsh: moved this file to a separate PR

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
elsh: ditto ^

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
elsh: ditto ^

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
elsh: ditto ^

Sources/MockoloFramework/Operations/ProtocolMapGenerator.swift
rudro: I agree, though in the 3 copies, there are 2 distinct uses. Two where we have an array and it's a lot like the GCD one and then one where the scanPath is scanning directories slowly and producing filenames one at a time and we dispatch them as they become available one by one. I suppose the make the API for the latter case and it would still work in the array case.

Tests/MockoloTestCase.swift
```diff
@@ -4,9 +4,9 @@ import MockoloFramework
 class MockoloTestCase: XCTestCase {
     var srcFilePathsCount = 1
     var mockFilePathsCount = 1
-
+
```
rudro: Is this file being intentionally being reformatted?

Sources/MockoloFramework/Operations/ProtocolMapGenerator.swift
elsh: Will address this on a separate pr as this is not related to swiftsyntax skeleton code.

Tests/MockoloTestCase.swift
elsh: Autoindent seems to have done this. PR #58 has test updates, so it'll be resolved there.

PR #58, opened 2019-11-25
Refactor models to be used with SwiftSyntax parser

elsh:


elsh:
> Not sure I can give you much feedback on this, I'm not familiar with how SwiftSyntax works. Maybe Murali could be more helpful?

already added him

Sources/MockoloFramework/Models/MethodModel.swift
```diff
@@ -16,85 +16,145 @@
...
+    
     init(_ ast: Structure, filepath: String, data: Data, processed: Bool) {
         // This will split func signature into name and the rest (params, return type). In case it's a generic func,
         // its type parameters will be in its substrctures (and < > are omitted in the func ast.name), so it will only
         // give the name part that we expect.
         var comps = ast.name.components(separatedBy: CharacterSet(arrayLiteral: ":", "(", ")")).filter {!$0.isEmpty}
         let nameString = comps.removeFirst()
         self.filePath = filepath
-
+
```
rudro: nit: In Xcode, could you go to `Preferences -> Text Editing -> Editing -> Automatically trim trailing whitespace` so we don't have these lines?

Sources/MockoloFramework/Models/MethodModel.swift
```diff
@@ -16,85 +16,145 @@
...
-        // Used to make the underlying function handler var name unique by providing args
-        // that can be appended to the name
-        self.signatureComponents = args.filter{ arg in !arg.isEmpty }
 
         // Sourcekit structure api doesn't provide info on throws/rethrows, so manually parse it here
         let suffixOffset = ast.nameOffset + ast.nameLength + 1
         let suffixLen = ast.offset + ast.length - suffixOffset
         if suffixLen > 0, suffixOffset > ast.bodyOffset - 1 {
-        let suffixPart = data.toString(offset: suffixOffset, length: suffixLen).trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
+            let suffixPart = data.toString(offset: suffixOffset, length: suffixLen).trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
```
rudro: nit: could we run `swiftformat` on all the files once so we can avoid these indentation fixes ?

Sources/MockoloFramework/Models/MethodModel.swift
```diff
@@ -16,85 +16,145 @@
...
     var type: Type
     var offset: Int64
     let length: Int64
     let accessControlLevelDescription: String
     let attributes: [String]
-    let staticKind: String
-    var genericTypeParams: [ParamModel] = []
-    var params: [ParamModel] = []
-    let handler: ClosureModel?
+    let genericTypeParams: [ParamModel]?
```
rudro: Seem asymmetric for this to be optional and the next one not. Can't the callers just pass in `[]` since that's what will happen later anyway?

Sources/MockoloFramework/Models/MethodModel.swift
```diff
@@ -16,85 +16,145 @@
...
+            }
+            return nil
+        }
+        if let genericParams = self.genericTypeParams {
+            let genericTypeNames = genericParams.map { $0.name.capitlizeFirstLetter + $0.type.displayName }
+            args.append(contentsOf: genericTypeNames)
+        }
+            
+        args.append(contentsOf: paramTypes.map(path: \.displayName))
+        var dtype = self.type.displayName
```
rudro: nit: `dType` or `displayType`

Sources/MockoloFramework/Models/MethodModel.swift
```diff
@@ -130,18 +182,23 @@ final class MethodModel: Model {
...
     func render(with identifier: String, typeKeys: [String: String]? = nil) -> String? {
         if processed {
             if isInitializer {
                 return nil
             }
-            
-            let ret = self.data.toString(offset: offset, length: length)
-            return ret
+
+            if let modelDescription = modelDescription {
```
rudro: This whole block is just `return modelDescription ?? self.data?.toString(offset: offset, length: length)`

Sources/MockoloFramework/Models/ParamModel.swift
```diff
@@ -25,21 +25,31 @@ final class ParamModel: Model {
...
     
     var modelType: ModelType {
         return .parameter
     }
+
+    init(label: String = "", name: String, typeName: String, isGeneric: Bool = false, isInitializer: Bool = false, offset: Int64, length: Int64) {
+        self.name = name.trimmingCharacters(in: .whitespaces)
+        self.type = Type(typeName.trimmingCharacters(in: .whitespaces))
+
+        let lbl = label.trimmingCharacters(in: .whitespaces)
```
rudro: nit: `label`

Sources/MockoloFramework/Models/TypeAliasModel.swift
```diff
@@ -69,6 +84,12 @@ final class TypeAliasModel: Model {
     }
     
     func render(with identifier: String, typeKeys: [String: String]? = nil) -> String? {
+        if processed {
+            if let modelDescription = modelDescription {
+                return modelDescription
+            }
```
rudro: so if it's processed but there no modelDescription, does that mean something is wrong ? Should there be an `else` here ?

Sources/MockoloFramework/Models/MethodModel.swift
elsh: I changed it to be optional because swiftsyntax treats also this as an optional property. Is there a more preferred way: handle an optional within this class vs. at the caller?

elsh: made it nonoptional

Sources/MockoloFramework/Models/MethodModel.swift
elsh: it's already set to that

PR #60, opened 2019-11-27
Add SwiftSyntax parser implementations

elsh:


Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -94,11 +94,13 @@ final class Entity {
         return attributes.filter {$0.contains(String.available)}
     }
 
-    static func model(name: String,
-                      label: String = "",
+    static func model(label: String = "",
```
elsh: this is just a placeholder function to be called from swiftsyntax parser; it will be replaced with real models after #58 is merged.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -8,7 +8,40 @@
 import Foundation
 import SwiftSyntax
 
-class EntityVisitor: SyntaxVisitor {
+extension SyntaxParser {
+    public static func parse(_ fileData: Data, path: String,
+                             diagnosticEngine: DiagnosticEngine? = nil) throws -> SourceFileSyntax {
+        // Avoid using `String(contentsOf:)` because it creates a wrapped NSString.
+        let source = fileData.withUnsafeBytes { buf in
+            return String(decoding: buf.bindMemory(to: UInt8.self), as: UTF8.self)
```
rmaz: `init?(data: Data, encoding: String.Encoding)` not simpler? docs are empty though, so hard to say if this ends up getting copied or not.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -21,23 +54,330 @@ class EntityVisitor: SyntaxVisitor {
...
+        return varmodels
+    }
+    
+    private func memberList(_ members: MemberDeclListSyntax, overrides: [String: String]?, acl: String, processed: Bool) -> ([String], [Model], Bool) {
+        var attributeList = [String]()
+        var memberList = [Model]()
+        var hasInit = false
+        
+        for m in members {
+            if let varMember = m.decl as? VariableDeclSyntax {
```
mkruber: Maybe a switch case will be easier to parse here.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -21,23 +54,330 @@ class EntityVisitor: SyntaxVisitor {
...
+                let ret = varModels(varMember, acl: acl, processed: processed)
+                memberList.append(contentsOf: ret)
+                if let attrDesc = varMember.attributes?.withoutTrivia().description {
```
mkruber: input to `append` can be a return value of a helper function as it is duplicated in multiple places.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -21,23 +54,330 @@ class EntityVisitor: SyntaxVisitor {
...
+                    attributeList.append(attrDesc.trimmingCharacters(in: .whitespacesAndNewlines))
```
mkruber: And `attributeList.append` can be at the end of the for loop too as it is common across all conditions with a different parameter. It can be further simplified by having a helper function which takes `m` and returns a <model, attribute> pair that can be appended in this loop.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -21,23 +54,330 @@ class EntityVisitor: SyntaxVisitor {
...
+        if let parents = node.inheritanceClause?.inheritedTypeCollection {
+            for p in parents {
+                if let pname = p.firstToken?.text {
+                    parentList.append(pname)
+                }
+            }
+        }
+        
+        var aclDesc = ""
+        if let mds = node.modifiers{
```
mkruber: swift-format?

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -21,23 +54,330 @@ class EntityVisitor: SyntaxVisitor {
...
+    func visit(_ node: ClassDeclSyntax) -> SyntaxVisitorContinueKind {
+        var aclDesc = ""
+        if let mds = node.modifiers {
+            aclDesc = acl(mds)
+        }
+        
+        let (attributes, members, hasInit) = memberList(node.members.members, overrides: nil, acl: aclDesc, processed: true)
+        
+        var attributeList = attributes
+        if let attrDesc = node.attributes?.withoutTrivia().description {
```
mkruber: can use helper.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
```diff
@@ -21,23 +54,330 @@ class EntityVisitor: SyntaxVisitor {
...
+            let desc = node.importTok.text + " " + ret
+            imports.append(desc)
+        }
+        return .visitChildren
+    }
+}
+
+extension Trivia {
+    private func metadata(with annotation: String, in val: String) -> AnnotationMetadata? {
+        if val.contains(annotation) {
```
mkruber: comment with a simple example may help code comprehensibility.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
elsh: In this case Data will be copied and then a new String instance is created with it. Also this is from one of swiftsyntax parse functions minus some overhead so would like to keep it as close to the original func as possible.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
elsh: this does more than just a case check, it casts as a vardeclsyntax which i can retrieve certain properties from.

Sources/MockoloFramework/Utils/SwiftSyntaxExtensions.swift
elsh: This is adding an array of model elements whereas it's just adding an element below.

PR #65, opened 2019-12-05
Optimization: Lazy model generation 

elsh:
Moved model creation to when actually needed

### DIFF
--- a/Sources/MockoloFramework/Models/ClassModel.swift
+++ b/Sources/MockoloFramework/Models/ClassModel.swift
@@ -46,7 +46,8 @@ final class ClassModel: Model {
         self.needInit = needInit
         self.initParams = initParams
         self.offset = offset
-        self.attribute = Set(attributes).joined(separator: " ")
+        let ret = attributes.filter {$0.contains(String.available)}
+        self.attribute = Set(ret).joined(separator: " ")
         self.accessControlLevelDescription = acl.isEmpty ? "" : acl + " "
         self.typealiasWhitelist = typealiasWhitelist
     }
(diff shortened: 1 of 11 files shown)


Sources/MockoloFramework/Models/ClassModel.swift
```diff
@@ -46,7 +46,8 @@ final class ClassModel: Model {
         self.needInit = needInit
         self.initParams = initParams
         self.offset = offset
-        self.attribute = Set(attributes).joined(separator: " ")
+        let ret = attributes.filter {$0.contains(String.available)}
```
benpious: This has been bothering me for a while; what is `ret`? "Return?" Can we use a more descriptive name like `availableAttributes`?

Sources/MockoloFramework/Models/MethodModel.swift
```diff
@@ -134,8 +134,9 @@ final class MethodModel: Model {
         assert(paramDecls.count == comps.count)
         
         let zippedParams = zip(paramDecls, comps)
-        self.params = zippedParams.map { (argAst: Structure, argLabel: String) -> ParamModel in
-            ParamModel(argAst, label: argLabel, offset: argAst.offset, length: argAst.length, data: data, isInitializer: ast.isInitializer)
+        self.params = zippedParams.map { (arg: (Structure, String)) -> ParamModel in
```
benpious: You probably don't need this. Swift generally only gives this diagnostic when it needs it to give the real error.

Sources/MockoloFramework/Models/ParamModel.swift
```diff
@@ -24,39 +24,39 @@ final class ParamModel: Model {
     var type: Type
     let label: String
     let isGeneric: Bool
-    let isInitializer: Bool
+    let inInit: Bool
```
benpious: What's the purpose of this change?

elsh: this description is more accurate than is*

Sources/MockoloFramework/Models/ClassModel.swift
elsh: yes

PR #66, opened 2019-12-18
Add a build/install script for distribution

elsh:
Distributing mockolo now requires shipping with lib_InternalSwiftSyntaxParser.dylib due to the use of the latest SwiftSyntax (from swift 5.1). This script handles build/install requirements.

### DIFF
--- a/install-script.sh
+++ b/install-script.sh
@@ -64,22 +64,22 @@ echo "OUTPUT FILE = ${OUTFILE}"
 
 
 echo "** Building..."
-cd $SRCDIR
+cd "$SRCDIR"
 rm -rf .build
 swift build -c release 
 
 cd .build/release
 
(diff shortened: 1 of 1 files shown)


install-script.sh
```diff
@@ -0,0 +1,85 @@
...
+
+echo "** Processing..."
+
+echo "SOURCE DIR = ${SRCDIR}"
+echo "TARGET = ${TARGET}"
+echo "DESTINATION DIR = ${DESTDIR}"
+echo "OUTPUT FILE = ${OUTFILE}"
+
+
+echo "** Building..."
```
uber-mahyar2: Say "** Clean & Build" ?

elsh: good idea

elsh:
cc @rudro @mkruber

PR #76, opened 2020-02-03
Fix for swift keyword named closure params

maxwellE:
Fixes an issue where generated mock code using  the name `for` as a closure argument to a function. These changes make sure to correctly escape `for` in this situation and with other closure keyword arguments

### DIFF
--- a/Sources/MockoloFramework/Templates/ClosureTemplate.swift
+++ b/Sources/MockoloFramework/Templates/ClosureTemplate.swift
@@ -32,6 +32,9 @@ func applyClosureTemplate(name: String,
             if argType.typeName.hasPrefix(String.autoclosure) {
                 return argName + "()"
             }
+            if argName.isSwiftKeyword {
+                return "`\(argName)`"
+            }
             return argName
         }
         handlerParamValsStr = zipped.joined(separator: ", ")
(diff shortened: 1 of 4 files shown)


Tests/TestNonSimpleVars/FixtureNonSimpleFuncs.swift
```diff
@@ -416,3 +416,60 @@ class NonSimpleFuncsMock: NonSimpleFuncs {
     }
 }
 """
+
+let forArgClosureFunc = """
+import Foundation
+
+/// \(String.mockAnnotation)
+protocol NonSimpleFuncs {
+func max(for: Int) -> (() -> Void)?
```
elsh: can you add a test case for 
func max(for key: Int) -> (() -> Void)?

Sources/MockoloFramework/Templates/ClosureTemplate.swift
```diff
@@ -32,6 +32,9 @@ func applyClosureTemplate(name: String,
             if argType.typeName.hasPrefix(String.autoclosure) {
                 return argName + "()"
             }
+            if argName.isSwiftKeyword {
```
elsh: add handling for an argument following a keyword (see my comment on test below for an example)

maxwellE: Will add these tests today

PR #77, opened 2020-02-05
Fix for `for` named closure params

maxwellE:
Fixes an issue where generated mock code using the name for as a closure argument to a function. These changes make sure to correctly escape for in this situation and with other closure keyword arguments

### DIFF
--- a/Sources/MockoloFramework/Templates/ClosureTemplate.swift
+++ b/Sources/MockoloFramework/Templates/ClosureTemplate.swift
@@ -32,10 +32,7 @@ func applyClosureTemplate(name: String,
             if argType.typeName.hasPrefix(String.autoclosure) {
                 return argName + "()"
             }
-            if argName.isSwiftKeyword {
-                return "`\(argName)`"
-            }
-            return argName
+            return argName.safeName
         }
         handlerParamValsStr = zipped.joined(separator: ", ")
     }
--- a/Sources/MockoloFramework/Templates/ClosureTemplate.swift
+++ b/Sources/MockoloFramework/Templates/ClosureTemplate.swift
@@ -30,7 +30,7 @@ func applyClosureTemplate(name: String,
         let zipped = zip(paramVals, paramTypes).map { (arg) -> String in
             let (argName, argType) = arg
             if argType.typeName.hasPrefix(String.autoclosure) {
-                return argName + "()"
+                return argName.safeName + "()"
             }
             return argName.safeName
         }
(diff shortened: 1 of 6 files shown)


Sources/MockoloFramework/Utils/StringExtensions.swift
```diff
@@ -70,6 +78,23 @@ extension String {
     var isThrowsOrRethrows: Bool {
         return self == .throws || self == .rethrows
     }
+
+    var isSwiftKeyword: Bool {
```
elsh: Instead of this, create var safeName: String { .. } in which it checks to see if self is contained in keywords and then return `self` if so.  Also move the keyword array out of StringExtension and make it a private (or fileprivate) constant. 

Then you can use safeName in the closureTemplate above.

Sources/MockoloFramework/Templates/ClosureTemplate.swift
```diff
@@ -32,6 +32,9 @@ func applyClosureTemplate(name: String,
             if argType.typeName.hasPrefix(String.autoclosure) {
                 return argName + "()"
             }
+            if argName.isSwiftKeyword {
```
elsh: remove this and use argName.safeName (see the comment below) everywhere in this block

Sources/MockoloFramework/Templates/ClosureTemplate.swift
```diff
@@ -32,7 +32,7 @@ func applyClosureTemplate(name: String,
             if argType.typeName.hasPrefix(String.autoclosure) {
                 return argName + "()"
             }
-            return argName
+            return argName.safeName
```
elsh: use safeName on line 33 also

Sources/MockoloFramework/Utils/StringExtensions.swift
```diff
@@ -70,6 +78,27 @@ extension String {
     var isThrowsOrRethrows: Bool {
         return self == .throws || self == .rethrows
     }
+
+    var safeName: String {
+        let kewords: [String] = [
```
elsh: declare this as a static let keywords array -- otherwise it'll create the list every time

Sources/MockoloFramework/Utils/StringExtensions.swift
```diff
@@ -43,6 +43,14 @@ extension String {
     static let `throws` = "throws"
     static let `rethrows` = "rethrows"
     static let `try` = "try"
+    static let `for` = "for"
```
elsh: Move these into an array and declare static let keywords

PR #80, opened 2020-02-09
Updates to override class members

elsh:


Sources/MockoloFramework/Models/VariableModel.swift
```diff
@@ -68,21 +68,29 @@ final class VariableModel: Model {
         self.filePath = filepath
     }
     
+    private func isGenerated(_ name: String) -> Bool {
+        return name.hasPrefix(.underlyingVarPrefix) || name.hasSuffix(.setCallCountSuffix) || name.hasSuffix(.callCountSuffix) || name.hasSuffix(.subjectSuffix) || name.hasSuffix("SubjectKind") || name.hasSuffix(.handlerSuffix)
```
NicoHinderling: how come "SubjectKind" isn't its own enum case?

elsh: ah yes, will update

Sources/MockoloFramework/Templates/ClassTemplate.swift
```diff
@@ -94,22 +93,6 @@ private func extraInitsIfNeeded(_ entities: [(String, Model)],
...
-                    if !matchingParams.isEmpty {
-                        matchingParamsFound = true
-                        break
-                    }
-                }
-            }
-            
-            if matchingParamsFound || extraInitParamCandidates.isEmpty {
-                needParamedInit = false
-            }
```
NicoHinderling: How come we don't need any of this anymore? Admittedly I don't have context on the role that `needParamedInit` plays, so I'm assuming it's related to that

elsh: extra init's are still generated for protocol mocks but not for class mocks.  this change handles that.

PR #82, opened 2020-02-12
Support members wrapped in #if macro

elsh:


FranzBusch:
@elsh Great improvement! I just quickly looked over this PR but we were facing a similar problem with `import` statements wrapped in `#if ` macros. All branches were picked up by mockolo and imported in the generated file. Would this PR also solve that and keep the imports in their respective macros?

elsh:
> @elsh Great improvement! I just quickly looked over this PR but we were facing a similar problem with `import` statements wrapped in `#if ` macros. All branches were picked up by mockolo and imported in the generated file. Would this PR also solve that and keep the imports in their respective macros?

Thanks @FranzBusch this particular PR doesn't handle that case.  But please file an issue on this site and we'll add support for that as well.

PR #86, opened 2020-02-19
Adds ability to add @testable import statements

maxwellE:
Adds the ability to add @testable import statements via the -i flag. This is useful for users who want to only add their mocks to a project's Test target and use the @testable import functionality to import the main projects target

The updated documentation reads as so:

` --testable-imports, -i  If set, @testable import statments will be added for each module name in this list.`

### DIFF
--- a/Sources/Mockolo/Executor.swift
+++ b/Sources/Mockolo/Executor.swift
@@ -32,6 +32,7 @@ class Executor {
     private var exclusionSuffixes: OptionArgument<[String]>!
     private var header: OptionArgument<String>!
     private var macro: OptionArgument<String>!
+    private var testableImports: OptionArgument<[String]>!
     private var annotation: OptionArgument<String>!
     private var concurrencyLimit: OptionArgument<Int>!
     private var useSourceKit: OptionArgument<Bool>!
@@ -97,6 +98,10 @@ class Executor {
                                 shortName: "-m",
                                 kind: String.self,
                                 usage: "If set, #if [macro] / #endif will be added to the generated mock file content to guard compilation.")
+        testableImports = parser.add(option: "--testable-imports",
+                                        shortName: "-i",
+                                        kind: [String].self,
+                                        usage: "If set, @testable import statments will be added for each module name in this list.")
         header = parser.add(option: "--header",
                                 shortName: "-h",
                                 kind: String.self,
@@ -158,6 +163,7 @@ class Executor {
         let header = arguments.get(self.header)
         let loggingLevel = arguments.get(self.loggingLevel) ?? 0
         let macro = arguments.get(self.macro)
+        let testableImports = arguments.get(self.testableImports) ?? []
         let shouldUseSourceKit = arguments.get(useSourceKit) ?? false
 
         do {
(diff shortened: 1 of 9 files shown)


Sources/MockoloFramework/Operations/Generator.swift
```diff
@@ -36,6 +36,7 @@ public func generate(sourceDirs: [String]?,
                      annotation: String,
                      header: String?,
                      macro: String?,
+                     testableImports: [String],
```
elsh: Since this is optional, make this an optional type to be consistent with other optional args.

Sources/MockoloFramework/Operations/OutputWriter.swift
```diff
@@ -37,8 +38,21 @@ func write(candidates: [(String, Int64)],
         break
     }
 
-    let importsSet = Set(importLines.map{$0.trimmingCharacters(in: .whitespaces)})
-    let importLineStr = importsSet.sorted().joined(separator: "\n")
+    var imports = importLines.compactMap { (importLine) -> String? in
```
elsh: Add a check if testableimports is nil or empty, and do this only if not; else it's converting to and from module and import lines redundantly.

Sources/MockoloFramework/Operations/OutputWriter.swift
```diff
@@ -37,8 +38,21 @@ func write(candidates: [(String, Int64)],
...
+        return extractModuleImport(importLine: importLine.trimmingCharacters(in: .whitespaces))
```
elsh: the extract function can live in stringextensions instead, then you can just call importLine.moduleName.

Tests/MockoloTestCase.swift
```diff
@@ -124,6 +124,7 @@ class MockoloTestCase: XCTestCase {
                       annotation: String.mockAnnotation,
                       header: header,
                       macro: "MOCK",
+                      testableImports: testableImports ?? [],
```
elsh: just pass in the arg; no need for ??

Sources/Mockolo/Executor.swift
```diff
@@ -158,6 +163,7 @@ class Executor {
         let header = arguments.get(self.header)
         let loggingLevel = arguments.get(self.loggingLevel) ?? 0
         let macro = arguments.get(self.macro)
+        let testableImports = arguments.get(self.testableImports) ?? []
```
elsh: no need for ??

Sources/MockoloFramework/Operations/OutputWriter.swift
```diff
@@ -36,9 +37,27 @@ func write(candidates: [(String, Int64)],
...
+    var importLineStr = ""
+    
+    if let testableImports = testableImports {
+        var imports = importLines.compactMap { (importLine) -> String? in
+            return importLine.moduleName
+        }
+        imports.append(contentsOf: testableImports)
+        importLineStr = Set(imports)
+            .sorted()
+            .map { importString -> String in
```
elsh: nit: arg name can be more clear, e.g. testableModuleName, rather than importString.

maxwellE:
Closed and reopened on a non master branch, #89

PR #91, opened 2020-02-27
Remove redundant var getters / Support rx subject overrides / Fix indentations

elsh:


Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -66,7 +104,7 @@ final class EntityNodeSubContainer {
 // Contains arguments to annotation
 // Ex. @mockable(typealias: T = Any; U = String; ...)
 struct AnnotationMetadata {
-    var typealiases: [String: String]?
+    var overrides: [String: String]?
```
benpious: Seems weird to have a struct with one optional param. Shouldn't the whole struct be optional?

Sources/MockoloFramework/Parsers/SourceKitExtensions.swift
```diff
@@ -34,29 +34,45 @@ extension Structure: EntityNode {
...
+        }
+        ret.overrides = args
 
+        return ret
+    }
+    
+    private func parseAnnotationArguments(key: Data, in extracted: Data) -> [String: String]? {
+        if let keyRange = extracted.range(of: key) {
+            let args = extracted[keyRange.endIndex...]
+            let argsStr = String(data: args, encoding: .utf8)?.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
```
benpious: nit: `CharacterSet` is unnecessary here.

Sources/MockoloFramework/Templates/ClassTemplate.swift
```diff
@@ -114,22 +125,21 @@ private func extraInitsIfNeeded(_ entities: [(String, Model)],
...
+            
         }.joined(separator: "\n")
         
+        
         initTemplate = """
-        \(accessControlLevelDescription)init(\(params)) {
-            \(paramsAssign)
-            \(String.doneInit) = true
-        }
+        \(String.spaces4)\(accessControlLevelDescription)init(\(params)) {
```
benpious: nit: I would write an extension on String that indents by N. Or something like `4.spaces` or `1.tabs`.

Sources/MockoloFramework/Utils/TypeParser.swift
```diff
@@ -303,34 +303,103 @@ public struct Type {
...
+                } else {
+                    return nil
+                }
+            }
+        }
+
+        if !vals.isEmpty {
+            var ret = vals.joined()
+            ret = ret.replacingOccurrences(of: ",", with: ", ")
+            ret = ret.replacingOccurrences(of: ",  ", with: ", ")
```
benpious: This is pretty subtle, it would be easy to be confused by this. How about something like `ret.split(",").map { $0.trimmingWhiteSpace }.joined(by: ", ") `?

Tests/MockoloTestCase.swift
```diff
@@ -132,6 +132,7 @@ class MockoloTestCase: XCTestCase {
                 let output = (try? String(contentsOf: URL(fileURLWithPath: self.dstFilePath), encoding: .utf8)) ?? ""
                 let outputContents = output.components(separatedBy:  CharacterSet.whitespacesAndNewlines).filter{!$0.isEmpty}
                 let fixtureContents = formattedDstContent.components(separatedBy: CharacterSet.whitespacesAndNewlines).filter{!$0.isEmpty}
+//                print(output)
```
benpious: Delete.

Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -25,13 +25,50 @@ struct ResolvedEntity {
...
+
+    var initParamCandidates: [Model] {
+        return sortedInitVars(in: uniqueModels.map{$0.1})
+    }
+
+    /// Returns models that can be used as parameters to an initializer
+    /// @param models The models (processed and unprocessed) of the current entity
+    /// @returns A list of init parameter models
+    private func sortedInitVars(`in` models: [Model]) -> [Model] {
+        let processed = models.filter {$0.processed && $0.canBeInitParam}
```
benpious: Can we be more explicit about what "processed" is?

Sources/MockoloFramework/Templates/ClassTemplate.swift
elsh: oh right, how about 4.ish ?

Sources/MockoloFramework/Utils/TypeParser.swift
elsh: good catch, i meant to do that before but haven't

Sources/MockoloFramework/Models/ParsedEntity.swift
elsh: It might have other things like module name etc down the road

PR #96, opened 2020-03-01
Add module name override support

elsh:
Add module name override support 
Clean up tests

/// @mockable(module: name = Foo) 
protocol Bar {
}

will now generate 
class BarMock: Foo.Bar {
}


elsh:
> Thanks @elsh for putting it together, I will try to test it a bit more later today. One thing I've noticed, is that the module prefix is applied to both, the mock and the protocol type.
> 
> I might be wrong, but I think the module prefix cannot be applied to the type definition, also, having it only as the protocol prefix make it more flexible - the mock can be included in the test target, or pulled aside and stored in a separated module (I prefer to store them in a dedicated module to make them reusable, and leveraged test modules).

You're right; updated pr

Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -102,23 +100,27 @@ final class EntityNodeSubContainer {
     }
 }
 
-// Contains arguments to annotation
-// Ex. @mockable(typealias: T = Any; U = String; ...)
+/// Contains arguments to annotation
+/// e.g. @mockable(module: name = Foo; typealias: T = Any; U = String; rx: barStream = PublishSubject)
```
benpious: `;` is a weird delimiter, shouldn't this be `,` to be valid Swift someday? 

As a more general point, this seems like it deserves a documentation change. I don't see any changes to .md files here.

benpious: Mockolo supports subclass mocks, right? You might want to rename this to just "prefix" so it's clear that you can reuse it for a scenario like this: 
class C {
    
   //// @mockable(prefix: "C")
    class D { 
     }

}

elsh: good idea, will do

elsh: > `;` is a weird delimiter, shouldn't this be `,` to be valid Swift someday?

The ; was simpler to parse as , can be used in a type argument for e.g. typealias  
 
> As a more general point, this seems like it deserves a documentation change. I don't see any changes to .md files here.

will add

marciniwanicki: Wouldn't AST give us enough information about the nested type as such the manual "prefix" attribute wouldn't be necessary in case of inner types? If we can infer it, then maybe it's better to use "module: name = <name>" or just use "module: <name>".

elsh: It would, but why not make this generic so it works for both inferrable and non-inferrable cases, in which case "prefix" makes sense. 

The format of the override argument needs to be consistent with other specifiers, e.g. 
@mockable(typealias: T = AnyObject) where 'typealias' is the keyword followed by a colon, and 'T' is the left operand to be overriden, and 'AnyObject' is the right operand to override it with. 

In case of "module", there could be other values to override in the future besides "prefix", such as "path".

PR #103, opened 2020-03-28
Update for SwiftSyntax 0.50200.0

akkyie:
Hi, I tried to update the SwiftSyntax API usage to get it working with the version 0.50200.0 and Xcode 11.4.

Here's changes I made and corresponding updates of SwiftSyntax API from its [Changelog](https://github.com/apple/swift-syntax/blob/master/Changelog.md):

##### 04bcf1b

> The walk method on syntax nodes has been removed.
> Instead, use the walk method on the SyntaxVisitor.
>
>     // Before 
>     tree.walk(&visitor)
>     // Now
>     visitor.walk(tree)

##### ac73789

> The protocols ExprSyntax, DeclSyntax, Syntax etc. have been removed

> For passing values of these types around, use the new type erasers ExprSyntax, DeclSyntax, Syntax etc. instead. To add computed properties or functions to all expression nodes, write an extension on ExprSyntaxProtocol. To add methods to all syntax nodes, extend SyntaxProtcol.

##### 5645416

> Downcasting can no longer be performed using the as operator. For downcasting use the as(_: SyntaxProtocol) method on any type eraser.

##### ce525e7

> SyntaxVisitor and SyntaxAnyVisitor are a class and no longer a protocol.
> Any structs conforming to the above protocols need to become classes. Implementing methods need to be marked override and, if necessary, any mutating keywords need to be removed.

As Travis doesn't have Xcode 11.4 environment yet, I'd appreciate if you try it on your local. Thanks!


elsh:
thanks for the change; tested on local machine and all looks good.

uhooi:
Works fine with Mockolo 1.2.0.
thx😊

PR #114, opened 2020-06-06
Support function arguments history captor

### LINKED ISSUE #111: Is there any way to assert function argument histories?
I would't like to assert function argument histories like as bellow.
```swift
protocol Foo {
    func fooFunc(_ arg: Int)
    func barFunc(_ arg1: Int, arg2: String)
}

let mock = FooMock()
mock.fooFunc(1)
mock.fooFunc(2)
mock.fooFunc(3)

XCTAssertEqual(mock.fooFuncValues, [1, 2, 3])

mock.barFunc(1, "A")
mock.barFunc(2, "B")
mock.barFunc(3, "C")

XCTAssertEqual(mock.barFuncValues, [(1, "A"), (2, "B"), (3, "C")])
```

Is there any way to achive this? And if I can't do currently, can I contribute to implement this feature?

andooown:
resolve #111 

This PR adds the function arguments captor.

Example:

```swift
/// @mockable(history: fooFunc = true)
protocol Foo {
    func fooFunc(val: Int)
}
```

This will generate the following mock output:

```swift
class FooMock: Foo {
    init() { }

    var fooFuncCallCount = 0
    var fooFuncArgValues = [Int]()
    var fooFuncHandler: ((Int) -> ())?
    func fooFunc(val: Int) {
        fooFuncCallCount += 1
        fooFuncArgValues.append(val)
        if let fooFuncHandler = fooFuncHandler {
            fooFuncHandler(val)
        }
    }
}
```

and also, the arguments captor is generated for all func by passing `--enable-args-history` args to `mockolo` command (it will ignore `history` annotations).

Currently, the arguments captor doesn't support closure variables.


Sources/MockoloFramework/Utils/StringExtensions.swift
```diff
@@ -56,6 +56,7 @@ extension String {
     static let `public` = "public"
     static let `open` = "open"
     static let initializer = "init"
+    static let argsHistorySuffix = "Values"
```
elsh, line 59: "ArgValues"

Sources/Mockolo/Executor.swift
```diff
@@ -208,6 +213,7 @@ class Executor {
                          to: outputFilePath,
                          loggingLevel: loggingLevel,
                          concurrencyLimit: concurrencyLimit,
+                         captureAllFuncArgsHistory: shouldCaptureAllFuncArgsHistory,
```
elsh, line 216: Can we rename this to "enableArgsHistory"?  
-Capture sounds a bit like a capture list in a closure.
-Also no need to specify "Func" as args are passed into init only once, so it's already implied that it's a func.

And move this argument up after declType; we want to put bool params close to each other.

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+
+final class ArgumentsHistoryModel: Model {
+    var name: String
+    var type: Type
+    var offset: Int64 = .max
+    let suffix: String
+    let capturableParamNames: [String]
```
elsh, line 8: Using 'capturable' here is ok as it's contained in this model.

Sources/MockoloFramework/Models/ClassModel.swift
```diff
@@ -55,7 +55,7 @@ final class ClassModel: Model {
         self.accessLevel = acl
     }
     
-    func render(with identifier: String, encloser: String, useTemplateFunc: Bool, useMockObservable: Bool) -> String? {
-        return applyClassTemplate(name: name, identifier: self.identifier, accessLevel: accessLevel, attribute: attribute, declType: declType, metadata: metadata, useTemplateFunc: useTemplateFunc, useMockObservable: useMockObservable, initParamCandidates: initParamCandidates, declaredInits: declaredInits, entities: entities)
+    func render(with identifier: String, encloser: String, useTemplateFunc: Bool, useMockObservable: Bool, captureAllFuncArgsHistory: Bool = false) -> String? {
```
elsh, line 58: Ditto ^ "enableArgsHistory"

Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -101,11 +101,12 @@ final class EntityNodeSubContainer {
 }
 
 /// Contains arguments to annotation
-/// e.g. @mockable(module: prefix = Foo; typealias: T = Any; U = String; rx: barStream = PublishSubject)
+/// e.g. @mockable(module: prefix = Foo; typealias: T = Any; U = String; rx: barStream = PublishSubject; history: bazFunc = true)
 struct AnnotationMetadata {
     var module: String?
     var typeAliases: [String: String]?
     var varTypes: [String: String]?
+    var historyCapturedFuncs: [String]?
```
elsh, line 109: To be consistent with the rest, mention the decl type first, so -> "funcsWithArgsHistory"

Sources/MockoloFramework/Parsers/SourceKitExtensions.swift
```diff
@@ -157,7 +160,7 @@ extension Structure: EntityNode {
             }
             
             if validated {
-                return MethodModel(element, encloserType: encloserType, filepath: filepath, data: data, processed: processed)
+                return MethodModel(element, encloserType: encloserType, filepath: filepath, data: data, historyCapturedFuncs: metadata?.historyCapturedFuncs ?? [], processed: processed)
```
elsh, line 163: Ditto^ -- funcsWithArgsHistory

Sources/MockoloFramework/Models/MethodModel.swift
```diff
@@ -81,7 +82,20 @@ final class MethodModel: Model {
...
-    
+    lazy var argsHistory: ArgumentsHistoryModel? = {
+        if isInitializer {
+            return nil
+        }
+
+        let ret = ArgumentsHistoryModel(name: name,
+                                        genericTypeParams: genericTypeParams,
+                                        params: params,
+                                        isHistoryAnnotated: historyCapturedFuncs.contains(name),
```
elsh, line 93: if it doesn't contain, it should return nil instead.

Sources/MockoloFramework/Templates/MethodTemplate.swift
```diff
@@ -44,9 +46,12 @@ extension MethodModel {
             return ""
         default:
             
-            guard let handler = handler else { return "" }
+            guard let handler = handler, let argsHistory = argsHistory else { return "" }
```
elsh, line 49: Check argsHistory / enableArgsHistory below instead

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
```diff
@@ -0,0 +1,49 @@
...
+        // Value contains closure is not supported.
+        let capturables = params.filter { !$0.type.hasClosure }
+        self.capturableParamNames = capturables.map(path: \.name)
+        self.capturableParamTypes = capturables.map(path: \.type)
+        
+        let genericTypeNameList = genericTypeParams.map(path: \.name)
+        self.type = Type.toArgumentsHistoryType(with: capturableParamTypes, typeParams: genericTypeNameList)
+    }
+    
+    func needsCaptureHistory(force: Bool) -> Bool {
```
elsh, line 30: rename -> func enable(force:)

Sources/MockoloFramework/Utils/TypeParser.swift
```diff
@@ -542,6 +542,30 @@ public final class Type {
...
+        let displayableParamTypes = params.compactMap { (subtype: Type) -> String? in
+            var processedType = subtype.processTypeParams(with: typeParams)
+            
+            if subtype.isInOut {
+                processedType = String(processedType.dropFirst((String.inout + " ").count))
+            }
+            if subtype.isIUO {
+                processedType = String(processedType.dropLast("!".count))
+            }
+
```
elsh, line 557: Can you add handling for generic type params?  See how they are handled in the Closure model (functionHandler).

Tests/TestArgumentsHistory/ArgumentsHistoryTests.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+
+class ArgumentsHistoryTests: MockoloTestCase {
+
```
elsh, line 4: Can you add a generic function test case?

Sources/Mockolo/Executor.swift
```diff
@@ -132,6 +133,9 @@ class Executor {
         useSourceKit = parser.add(option: "--use-sourcekit",
                              kind: Bool.self,
                              usage: "Whether to use SourceKit for parsing. By default it will use SwiftSyntax.")
+        enableArgsHistory = parser.add(option: "--enable-args-history",
+                                       kind: Bool.self,
+                                       usage: "Whether generated mock contains function args history capturing. If set, `history` annotation will be ignored (default = false).")
```
elsh, line 138: "Whether to enable args history for all functions (default = false).  To enable history per function, use the 'history' keyword in the annotation argument. "

Sources/Mockolo/Executor.swift
andooown: > Can we rename this to "enableFuncArgsHistory" or "enableArgsHistory"? Capture sounds a bit like a capture list in a closure.
Sure. I will rename it to `enableArgsHistory`!

> And move this argument up after declType; we want to put bool params close to each other.
OK 😃

Sources/Mockolo/Executor.swift
andooown: Fixed this and README in 2164945.

Sources/MockoloFramework/Utils/StringExtensions.swift
andooown: Fixed this in d314b82.

Sources/Mockolo/Executor.swift
andooown: I moved this after `useMockObservable` and renamed in d621414.

Sources/MockoloFramework/Models/ClassModel.swift
andooown: Also fixed in d621414.

Sources/MockoloFramework/Models/ParsedEntity.swift
andooown: Fixed this is 16ee27f.

Sources/MockoloFramework/Parsers/SourceKitExtensions.swift
andooown: Also fixed this is 16ee27f.

Sources/MockoloFramework/Models/MethodModel.swift
andooown: Whether the function is annotated or not, the arguments captor must be rendered by `ArgumentsHistoryModel` if `--enable-args-history` is passed.
And it can know in `render(with:encloser:useTemplateFunc:useMockObservable:enableFuncArgsHistory:)`.

So it should return model in this time, isn't it?

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
andooown: Maybe `enabled(force:)` is better than `enable(force:)`?
Because this return the state.

Tests/TestArgumentsHistory/ArgumentsHistoryTests.swift
andooown: Added in a26573b and b35d242.

Sources/MockoloFramework/Templates/MethodTemplate.swift
andooown: Fixed.

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
```diff
@@ -0,0 +1,49 @@
...
+        return .class
+    }
+
+    init(name: String, genericTypeParams: [ParamModel], params: [ParamModel], isHistoryAnnotated: Bool, suffix: String) {
+        self.name = name + .argsHistorySuffix
+        self.suffix = suffix
+        self.isHistoryAnnotated = isHistoryAnnotated
+
+        // Value contains closure is not supported.
+        let capturables = params.filter { !$0.type.hasClosure }
```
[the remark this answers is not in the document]
andooown: Filter out subscript and static func handling with test cases.
It is worked 👍

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
andooown: On second thought, `enable(force:)` was more natural.
I will fix it.

andooown: Fixed in b35d242.

Sources/MockoloFramework/Utils/TypeParser.swift
andooown: > Can you add handling for generic type params and overloaded funcs with the same name? See how they are handled in the Closure model (function handler).

Handled in df1a3bd and added test cases was passed ✅ 

> Re: overloaded funcs, we can, for now, allow args of all overloaded funcs to be saved if enabled via the annotation 'history' argument. The name of the argumenthistorymodel will need to be unique in this case -- you can follow the naming of the function handler closure model to be consistent.

This means that users must annotate like as following?
```
/// @mockable(history: fooFuncValString = true)
protocol Foo {
    func fooFunc(val: Int)
    func fooFunc(val: String)  // To enable history only for it.
}
```

It is seems confusing to me 😢 
In this case, I think that capturing history of all of fooFuncs isn't bad.

andooown:
I've addressed the points you raised 😼 
I want you to check again!

Tests/TestArgumentsHistory/FixtureArgumentsHistory.swift
```diff
@@ -0,0 +1,382 @@
...
+
+let argumentsHistoryOverloadedCaseMock = """
+class FooMock: Foo {
+    init() { }
+
+    var fooFuncCallCount = 0
+    var fooFuncHandler: (() -> ())?
+    func fooFunc() {
+        fooFuncCallCount += 1
+
```
elsh, line 201: where's argValues for this one?

Sources/MockoloFramework/Utils/TypeParser.swift
elsh: Yeah, to simplify, we can allow all overloaded funcs to have args history for now.

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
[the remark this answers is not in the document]
elsh: Cool. Another point: move !capturables.isEmpty check to the beginning, and also filter out if an arg type is @autoclosure; return nil if empty (this init becomes init?), so the rest don't have to be executed in such case.

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
```diff
@@ -0,0 +1,49 @@
...
+        self.type = Type.toArgumentsHistoryType(with: capturableParamTypes, typeParams: genericTypeNameList)
+    }
+    
+    func enable(force: Bool) -> Bool {
+        return (force || isHistoryAnnotated) && !capturableParamNames.isEmpty
+    }
+    
+    func render(with identifier: String, encloser: String, useTemplateFunc: Bool = false, useMockObservable: Bool = false, enableFuncArgsHistory: Bool) -> String? {
+        guard enable(force: enableFuncArgsHistory) else {
+            return ""
```
elsh, line 36: return nil

Tests/TestArgumentsHistory/FixtureArgumentsHistory.swift
```diff
@@ -0,0 +1,382 @@
...
+let argumentsHistorySimpleCaseMock = """
+class FooMock: Foo {
+    init() { }
+
+    var fooFuncCallCount = 0
+    var fooFuncHandler: (() -> ())?
+    func fooFunc() {
+        fooFuncCallCount += 1
+
```
elsh, line 88: argValues?

Tests/TestArgumentsHistory/FixtureArgumentsHistory.swift
```diff
@@ -0,0 +1,382 @@
...
+let argumentsHistoryHandlerCaseMock = """
+class FooMock: Foo {
+    init() { }
+
+    var fooFuncCallCount = 0
+    var fooFuncHandler: ((() -> Int) -> ())?
+    func fooFunc(handler: () -> Int) {
+        fooFuncCallCount += 1
+
```
elsh, line 339: ditto^

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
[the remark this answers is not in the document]
andooown: NOTE: Filter out subscript in here.

https://github.com/andooown/mockolo/blob/2f1da75b2ebf0a753536e0dd72990a6492bf99dd/Sources/MockoloFramework/Models/MethodModel.swift#L91-L103

Sources/MockoloFramework/Models/MethodModel.swift
elsh: That's correct. This is fine then.

Tests/TestArgumentsHistory/FixtureArgumentsHistory.swift
andooown: `fooFunc()`doesn't have any arguments.

Tests/TestArgumentsHistory/FixtureArgumentsHistory.swift
andooown: Same as above.
`fooFunc(handler: () -> Int)` doesn't have any capturable arguments.

Tests/TestArgumentsHistory/FixtureArgumentsHistory.swift
andooown: ditto.

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
andooown: Fixed it in f592f3e.

Sources/MockoloFramework/Models/ArgumentsHistoryModel.swift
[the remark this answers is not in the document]
andooown: Addressed in 621e974 and add autoclosure test cases.

andooown:
mmm... it seems that Travis didn't work?

Tests/TestArgumentsHistory/FixtureArgumentsHistory.swift
elsh: ah, overlooked. this looks good then.

PR #118, opened 2020-07-28
[ReturnSelf] Allows Protocols to Return Self

tinder-maxwellelliott:
Currently Mockolo is broken for Protocols that return Self. These changes ensure that we return the mock class type when we encounter a Self return type:

Example protocol that breaks Mockolo currently

```
/// @mockable
protocol NonSimpleFuncs {
    @discardableResult
    func returnSelf() -> Self
}
```

### DIFF
--- a/Sources/MockoloFramework/Models/MethodModel.swift
+++ b/Sources/MockoloFramework/Models/MethodModel.swift
@@ -123,6 +123,7 @@ final class MethodModel: Model {
     init(name: String,
          typeName: String,
          kind: MethodKind,
+         encloser: String?,
          encloserType: DeclType,
          acl: String,
          genericTypeParams: [ParamModel],
@@ -135,7 +136,7 @@ final class MethodModel: Model {
          modelDescription: String?,
          processed: Bool) {
         self.name = name.trimmingCharacters(in: .whitespaces)
-        self.type = Type(typeName.trimmingCharacters(in: .whitespaces))
+        self.type = Type(typeName.trimmingCharacters(in: .whitespaces), encloser: encloser)
         self.suffix = throwsOrRethrows
         self.offset = offset
         self.length = length
--- a/Sources/MockoloFramework/Models/MethodModel.swift
+++ b/Sources/MockoloFramework/Models/MethodModel.swift
@@ -40,7 +40,6 @@ final class MethodModel: Model {
     let suffix: String
     let kind: MethodKind
     let funcsWithArgsHistory: [String]
-    let encloser: String?
     var modelType: ModelType {
         return .method
     }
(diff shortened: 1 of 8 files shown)


Tests/TestNonSimpleCases/FixtureNonSimpleFuncs.swift
```diff
@@ -514,3 +514,33 @@ class NonSimpleFuncsMock: NonSimpleFuncs {
...
+
+import Foundation
+
+
+class NonSimpleFuncsMock: NonSimpleFuncs {
+    init() { }
+
+    var returnSelfCallCount = 0
+    var returnSelfHandler: (() -> (NonSimpleFuncsMock))?
+    func returnSelf() -> NonSimpleFuncsMock {
```
elsh, line 538: The return type here should still be Self as the encloser type is not final.  We could make it final but I've seen cases where some mocks are subclassed for customization so will need to audit first.

Sources/MockoloFramework/Models/MethodModel.swift
```diff
@@ -123,6 +125,7 @@ final class MethodModel: Model {
     init(name: String,
          typeName: String,
          kind: MethodKind,
+         encloser: String?,
```
elsh, line 128: No need to pass in encloser here; closuremodel is created lazily during rendering (see render(with:) function in this file), where the encloser is passed in from the mock class. You can change the 'handler' var to a function with an encloser parameter.

Sources/MockoloFramework/Parsers/ViaSwiftSyntax/SwiftSyntaxExtensions.swift
```diff
@@ -146,7 +146,7 @@ extension MemberDeclListItemSyntax {
         return modifiers?.acl ?? ""
     }
     
-    func transformToModel(with encloserAcl: String, declType: DeclType, metadata: AnnotationMetadata?, processed: Bool) -> (Model, String?, Bool)? {
+    func transformToModel(with encloserAcl: String, declType: DeclType, metadata: AnnotationMetadata?, processed: Bool, encloser: String?) -> (Model, String?, Bool)? {
```
elsh, line 149: see comment above^

Tests/TestNonSimpleCases/FixtureNonSimpleFuncs.swift
```diff
@@ -514,3 +514,33 @@ class NonSimpleFuncsMock: NonSimpleFuncs {
...
+
+class NonSimpleFuncsMock: NonSimpleFuncs {
+    init() { }
+
+    var returnSelfCallCount = 0
+    var returnSelfHandler: (() -> (NonSimpleFuncsMock))?
+    func returnSelf() -> Self {
+        returnSelfCallCount += 1
+        if let returnSelfHandler = returnSelfHandler {
+            return returnSelfHandler()
```
elsh, line 541: this probably requires casting to Self

elsh:
thanks for the change

Issue #131, opened 2020-12-13
Avoid writing to destination file if content is unchanged


ffittschen:
We recently investigated our build times and discovered that `mockolo` [is always writing](https://github.com/uber/mockolo/blob/7439d3b6f40f2169752825296fd5712cdd137ae2/Sources/MockoloFramework/Operations/OutputWriter.swift#L44) to the destination file even though the content did not change. This behavior will trigger Xcode to rebuild modules in which no source code was changed (and all of the modules that depend on this module). To improve this behavior, I'd suggest to check if the destination file already exists and only write the file if the content is unchanged.

To test the current behavior, I executed these commands:
```bash
stat Sources/Mocks.generated.swift
mockolo -s Sources/ -d Sources/Mocks.generated.swift -m DEBUG
stat Sources/Mocks.generated.swift
```

Looking at other tools that generate code, they also avoid to write the destination file if the content is unchanged. (See [SwiftGen](https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGen/Commander/OutputDestination.swift#L32-L44) or [R.swift](https://github.com/mac-cain13/R.swift/blob/master/Sources/RswiftCore/RswiftCore.swift#L180-L188))

If you accept this suggestion, I'd be happy to open a PR for this.

elsh:
Sorry for a delay. Left a comment on your PR.

PR #132, opened 2020-12-13
Avoid writing to destination file if content is unchanged

ffittschen:
Resolves: #131 

As proposed in #131, this PR adapts the `write` method to only write the destination file if its content did not change. ~In the [equivalent PR](https://github.com/uber/needle/pull/382) for `needle` I was able to add a unit test for the file write behavior, but `mockolo` doesn't depend on https://github.com/uber/swift-common, so I wasn't able to leverage the `UnitTestLogger`. Do you have any other ideas of how to test the file write behavior?~

**Update:**
I also added a test that compares the [`.creationDate`](https://developer.apple.com/documentation/foundation/fileattributekey/1418187-creationdate) and [`.modificationDate`](https://developer.apple.com/documentation/foundation/fileattributekey/1410058-modificationdate) when calling `write` repeatedly. If you have any suggestions for a better location of this test, please let me know.

### DIFF
--- a/Tests/TestFuncs/TestBasicFuncs/FuncBasicTests.swift
+++ b/Tests/TestFuncs/TestBasicFuncs/FuncBasicTests.swift
@@ -1,4 +1,5 @@
 import Foundation
+import XCTest
 
 class BasicFuncTests: MockoloTestCase {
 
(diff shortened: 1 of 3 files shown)


tinder-maxwellelliott (approved):

Tests/TestFuncs/TestBasicFuncs/FuncBasicTests.swift
```diff
@@ -48,5 +49,24 @@ class BasicFuncTests: MockoloTestCase {
                dstContent: simpleMockFuncMock,
                useTemplateFunc: true)
     }
+
+    func testFileWriteBehavior() throws {
```
elsh, line 53: Can you create a new dir, Tests/TestOutputChange and move this to it?

elsh (commented):
Thanks for making this change. Please address the test location and we'll merge this PR.

ffittschen:
@elsh Thanks for the review, I just updated the PR and moved the test to the new location.

elsh (approved):

Issue #110, opened 2020-05-09
Support for Combine


vsanthanam:
Add support for annotation attributes for Combine subjects and publishers so they too can be easily mocked.

elsh:
@vsanthanam  Per discussion, could you add examples here?

vsanthanam:
When mockolo is mocking a property, is does something special if the property is an `Observable<Element>`

The generated mock will not crash you try to access the property before assignment, and is instead automatically backed `PublishSubject`. An additional ivar is created to allow the consumer to add values to the sequence, and additional options exist to control what kind of subject is used to back the `Observable` (e.g. replay, behavior, etc.)

I think we should add the same functionality when a property is of type `AnyPublisher` (Apple's equivalent of Observable — a ready only, type erased value that the consumer can subscribe to but cannot add values to)

We should also back these types automatically with a combine subject, and provide some controls over what kind of subjects could be used. Unlike Observables which are generics specialized against a single type, Publishers are specialized with two types (the value to emit, as well as the error), so this could add some complexity. On the plus side though, Combine has much fewer features than RxSwift as of now, and only there are only two subject types to choose from.

For example:

```
/// @mockable
protocol StateManaging {
   var myState: AnyPublisher<State, Error>
}
```

Could generate the following mock:

```
class StateManagingMock: StateManaging {
   var myState: AnyPublisher<State, Error> {
      myStateSubject.eraseToAnyPublisher() // the Combine equivalent of `.asObservable()` to perform type erasure of the mutable interface.
   }
   
   var myStateSubject = PassthroughSubject<State, Error>()
}
```

Combine really only has two subjects: PassthroughSubject (the equivalent of PublishSubject, initialized with no arguments), and CurrentValueSubject (the equivalent of BehaviorSubject, initialized with one value to start the sequence), so you'd have full coverage by supporting just those two.

PR #155, opened 2021-08-06
Support for mocking Combine's AnyPublisher

milettal:
Addresses: #110 

Allows developers to mock AnyPublishers in a few ways, either with a `subject` or a `published` property specifier.
i.e.
```
/// @mockable(subject: myIntPublisher = CurrentValueSubject;)(published: myStrPublisher = myStr;)
public protocol Foo {
    var myIntPublisher: AnyPublisher<Int, Never> { get }
    var myStrPublisher: AnyPublisher<String, Never> { get } 
}
```

results in

```
public class FooMock: Foo {
    public init() { }
    public init(myStr: String = "") {
        self.myStr = myStr
    }


    public var myIntPublisher: AnyPublisher<Int, Never> { return self.myIntPublisherSubject.eraseToAnyPublisher() }
    public private(set) var myIntPublisherSubject = CurrentValueSubject<Int, Never>(0)

    public private(set) var myStrSetCallCount = 0
    @Published public var myStr: String = "" { didSet { myStrSetCallCount += 1 } }

    public var myStrPublisher: AnyPublisher<String, Never> { return self.$myStr.setFailureType(to: Never.self).eraseToAnyPublisher() }
}

```

The published property also works if the property is already defined in the protocol/class or somewhere in the inheritance tree.

```
/// @mockable(published: myStrPublisher = myStr;)
public protocol Foo {
    var myStr: String { get }
    var myStrPublisher: AnyPublisher<String, Never> { get } 
}
```

Open to any and all feedback.


Tests/TestCombine/FixtureCombine.swift
```diff
@@ -0,0 +1,194 @@
+ import MockoloFramework
+
+ let combineProtocol = """
+/// \(String.mockAnnotation)(subject: dictionaryPublisher = CurrentValueSubject; myPublisher = PassthroughSubject; noDefaultSubjectValue = CurrentValueSubject)
```
EspressoCup, line 4: A better parameter name might be "combine" to be distinguished from "rx" which also allows Subjects.  The same param can be used to denote @Published as well, like so: 

`@mockable(combine: foo = CurrentValueSubject; bar = @Published)`

Tests/TestCombine/FixtureCombine.swift
```diff
@@ -0,0 +1,194 @@
...
+    public var dictionaryPublisher: AnyPublisher<Dictionary<String, String>, Never> { return self.dictionaryPublisherSubject.eraseToAnyPublisher() }
+    public private(set) var dictionaryPublisherSubject = CurrentValueSubject<Dictionary<String, String>, Never>(Dictionary<String, String>())
+
+    public var noDefaultSubjectValue: AnyPublisher<CustomType, Never> { return self.noDefaultSubjectValueSubject.eraseToAnyPublisher() }
+    public private(set) var noDefaultSubjectValueSubject = PassthroughSubject<CustomType, Never>()
+}
+"""
+
+let combinePublishedProtocol = """
+/// \(String.mockAnnotation)(published: myStringPublisher = myString; myIntPublisher = myInt; myCustomTypePublisher = myCustomType; myNonOptionalPublisher = myNonOptional)
```
EspressoCup, line 28: See the comment on the parameter name. Instead of a user needing to provide a var name to map to, it might be a better approach to autogenerate a unique name, like so:
 
`@mockable(combine: foo = @Published)` 
-> generates `@Published var publishedFoo // or to be more unique, atPublishedFoo, published_Foo, etc`.

This is unless there's a need to map a publisher var to a specific non-publisher var name; if that's the case, how common is it and how are they used in tests?

EspressoCup: Thanks for the explanation. That makes sense. A follow up question then is, if a publisher var doesn't have a matching non-published var declared in the protocol (or its parent protocols), would it actually need one (how is the publisher fired without it in non-test code?) or would it suffice to just provide a default PassthroughSubject?   

Re: annotation, it could be`@mockable(combine: foo = CurrentValueSubject; barPublisher = @Published bar)` to be more consistent with 'rx'.  Lmk your thoughts.

EspressoCup: If there's no matching var declared, then maybe it should be treated as an "incorrect" case and we shouldn't generate one at all (PassthroughSubject or not).

milettal: I updated the implementation to match what we discussed here.

> If there's no matching var declared, then maybe it should be treated as an "incorrect" case and we shouldn't generate one at all (PassthroughSubject or not).

I decided to generate the PassthroughSubject anyway in this case. If we didn't, the user would end up with the typical variable generation. Roughly:

```
///mockable(combine: myIntPublisher = @Published myInt)
protocol Foo {
    var myIntPublisher: AnyPublisher<Int, Never> { get }
}

class FooMock: Foo {
    init() { }
    init(myIntPublisher: AnyPublisher<Int, Never>) {
        self._myIntPublisher = myIntPublisher
    }

    var _myIntPublisher: AnyPublisher<Int, Never>! 

    var myIntPublisher: AnyPublisher<Int, Never> { _myIntPubilsher }
}
```

It seems more developer friendly to always give them a PassthroughSubject at the very least. Less setup on their part.

Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -99,6 +99,25 @@ final class EntityNodeSubContainer {
     }
 }
 
+public enum CombineSubjectType {
```
EspressoCup, line 102: this and enum below can be merged into one, e.g. 

```
public enum CombineType { 
 case passthroughSubject
 case currentValueSubject 
 case published(propertyWrapper: String, name: String) // or case property(wrapper, name)
}

File: Sources/MockoloFramework/Models/ParsedEntity.swift [hunk 4]
```diff
@@ -107,6 +126,8 @@ struct AnnotationMetadata {
     var varTypes: [String: String]?
     var funcsWithArgsHistory: [String]?
     var modifiers: [String: Modifier]?
+    var combineSubjectTypes: [String: CombineSubjectType]?
```
EspressoCup, line 129: one var of [String: CombineType] -- see comment on enum Combine-

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -685,53 +686,92 @@ extension Trivia {
...
+                modifiers[tuple.key] = modifier
             }
-            return ret
+            ret.modifiers = modifiers
         }
-        return nil
+        return ret
+    }
+
+    private func parseArguments(_ argsStr: String, identifier: String) -> [String: String]? {
```
EspressoCup, line 765: nice refactoring!

Sources/MockoloFramework/Models/VariableModel.swift
```diff
@@ -70,18 +75,18 @@ final class VariableModel: Model {
                 return prefix + modelDescription
             }
 
-            if let ret = self.data?.toString(offset: self.offset, length: self.length) {
```
EspressoCup, line 73: thanks for removing this

Sources/MockoloFramework/Operations/UniqueModelGenerator.swift
```diff
@@ -70,3 +72,33 @@ private func generateUniqueModels(key: String,
     
     return ResolvedEntityContainer(entity: resolvedEntity, paths: paths, imports: pathToContentList)
 }
+
+private func combinePostLookup(models: [Model]) {
```
EspressoCup, line 76: This should be called from ClassModel.applyClassTemplate in the beginning, to ensure all models are unique including the 'processed' models (that can be passed down from mock input files). See how typealias collisions are handled there for example.

Tests/TestCombine/CombineTests.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+class CombineTests: MockoloTestCase {
+
+    func testCombine() {
```
EspressoCup, line 5: Could you add a test with param `mockContent`? See `testSimpleInitParams` for an example.

EspressoCup (commented):
Thanks for the updates and cleanup! Left comments.

Tests/TestCombine/FixtureCombine.swift
EspressoCup: Makes sense, agreed.

Sources/MockoloFramework/Models/ParsedEntity.swift
milettal: Good call

Sources/MockoloFramework/Operations/UniqueModelGenerator.swift
milettal: Moved it + renamed

Tests/TestCombine/CombineTests.swift
milettal: Test added. Please check to ensure it's testing what you were looking for.

Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -99,14 +99,32 @@ final class EntityNodeSubContainer {
...
+            return .currentValueSubject
+        case .property:
+            return ""
+        }
+    }
+}
+
 /// Contains arguments to annotation
-/// e.g. @mockable(module: prefix = Foo; typealias: T = Any; U = String; rx: barStream = PublishSubject; history: bazFunc = true, modifiers: someVar = weak)
+/// e.g. @mockable(module: prefix = Foo; typealias: T = Any; U = String; rx: barStream = PublishSubject; history: bazFunc = true; modifiers: someVar = weak; combine: somePublisher = CurrentValuePublisher; otherPublisher = @Published otherProperty)
```
EspressoCup, line 120: s/CurrentValuePublisher/CurrentValueSubject

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -372,10 +372,10 @@ extension VariableDeclSyntax {
                                          isStatic: isStatic,
                                          canBeInitParam: potentialInitParam,
                                          offset: v.offset,
-                                         length: v.length,
-                                         overrideTypes: overrides,
-                                         customModifiers: customModifiers,
+                                         overrideTypes: metadata?.varTypes,
```
EspressoCup, line 375: side note: we'll need to rename this now to rxTypes

Tests/TestCombine/CombineTests.swift
EspressoCup: yup looks good

EspressoCup (approved):
lgtm; thanks for adding this feature!

Sources/MockoloFramework/Models/ParsedEntity.swift
EspressoCup: Maybe just ComebineType instead of CombineSubjectType? The latter seems limited to just the first 2 cases.

PR #165, opened 2021-10-05
Fix Swift Argument Parser

uhooi:
## Overview

Replace `exact` with `from`.

ref: https://github.com/apple/swift-package-manager/blob/ce50cb0de101c2d9a5742aaf70efc7c21e8f249b/Documentation/PackageDescription.md#methods-4

> /// Specifying exact version requirements are not recommended as
> /// they can cause conflicts in your dependency graph when multiple other packages depend on a package.
> /// As Swift packages follow the semantic versioning convention,
> /// think about specifying a version range instead.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,7 +3,7 @@ import PackageDescription
 
 var dependencies: [Package.Dependency] = [
     .package(url: "https://github.com/apple/swift-tools-support-core.git", .exact("0.2.3")),
-    .package(url: "https://github.com/apple/swift-argument-parser", .exact("1.0.1")),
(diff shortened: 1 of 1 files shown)


EspressoCup:
@uhooi Feel free to cut a new tag/release (see the past ones for example); make sure to bump the version first at https://github.com/uber/mockolo/blob/master/Sources/MockoloFramework/Version.swift (e.g. 1.6.1 -> 1.6.2)

uhooi:
@EspressoCup 
1.6.2 released!!
https://github.com/uber/mockolo/releases/tag/1.6.2

PR #173, opened 2022-01-13
Update array parsing strategy

### LINKED ISSUE #172: Bug: Specifying multiple mock file paths is broken
Prior to 1.5.0, it was possible to specify multiple mock file paths with the `-mocks` argument separated by a space, e.g.

```
mockolo -s Foo/Sources -mocks Bar/Sources/Mocks.generated.swift Baz/Sources/Mocks.generated.swift -d Foo/Sources/Mocks.generated.swift
```

With the switch to the Swift Argument Parser in 1.5.0, this doesn't seem to be possible anymore. When running `mockolo` directly in Xcode and adding a breakpoint in [Executor.swift#L173](https://github.com/uber/mockolo/blob/94a78fcc70f3c455d103b6bbce3ff068b8110941/Sources/Mockolo/Executor.swift#L173), the `mockFilePaths` will be an array with 1 value of `Bar/Sources/Mocks.generated.swift Baz/Sources/Mocks.generated.swift` instead of an array with 2 values `Bar/Sources/Mocks.generated.swift` and `Baz/Sources/Mocks.generated.swift`. This bug not only applies to the `-mocks` option, but all options that have _"[...] (separated by a comma or a space) [...]"_ in their help text.

To restore the previous behavior, we could make use of the [`.upToNextOption`](https://github.com/apple/swift-argument-parser/blob/9e1448215699a8757d9729450e4b7a1951d8af2a/Sources/ArgumentParser/Parsable%20Properties/Option.swift#L281) parsing strategy in the `@Option`.

I'd be happy to create a PR for this, but as there are also other `[String]` options, I thought I'd clarify it with you first to have a consistent usage for all array options.

ffittschen:
Resolves #172 

This PR updates the parsing strategy for all `[String]` options to `upToNextOption`, which will restore the functionality of passing multiple strings separated by space to via one option, e.g. `-mock path/to/foo path/to/bar`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,31 +89,79 @@ Use --help to see the complete argument options.
 
 OVERVIEW: Mockolo: Swift mock generator.
 
-USAGE: mockolo <options>
+USAGE: mockolo [<options>] --destination <destination>
 
 OPTIONS:
-  --allow-set-callcount     If set, generated *CallCount vars will be allowed to set manually. 
-  --annotation, -a          A custom annotation string used to indicate if a type should be mocked (default = @mockable).
-  --concurrency-limit, -j   Maximum number of threads to execute concurrently (default = number of cores on the running machine).
-  --custom-imports, -c      If set, custom module imports will be added to the final import statement list.
-  --destination, -d         Output file path containing the generated Swift mock classes. If no value is given, the program will exit.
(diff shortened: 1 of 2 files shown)


EspressoCup (approved):

EspressoCup:
thanks for the fix!

PR #186, opened 2022-04-25
Disable combine parameter overloading via a flag

maxwellE:
Allows consumers to disable the default behavior of declaring and initializing combine streams in mocks


EspressoCup:
Lgtm

Issue #182, opened 2022-03-28
xcode 13.3 - fatal error - SwiftSyntax.ParserError 1


SredzinskiPatryk:
`MockoloFramework/SourceParser.swift:103: Fatal error: The operation couldn’t be completed. (SwiftSyntax.ParserError error 1.)`

I do receive that error after updating my Xcode to 13.3. I was working previously on 13.2.1

ManuC2001:
same issue with brew installation.

 It's probably linked to the swift 5.6 compatibility (cf. Package.swift line 9)

I rebuild it from source code (option 4 in the installation doc) and removed this line in the Package.swift and solved the issue.

uhooi:
What version of Mockolo is installed by Homebrew?
I am wondering if `1.7.0` also causes an error.

psredzinski:
Yes it fails on 1.7.0

bezoadam:
I have version `1.7.0` and have same issue with Xcode 13.3

entaku0818:
I have same issue with Xcode 13.3.
Mockolo was installed by Mint.


```Mintfile
uber/mockolo@1.7.0
```

Below is the build log.

```
/Applications/Xcode.app/Contents/Developer/usr/bin/make generate-mocks-with-mockolo
mint run mockolo --sourcedirs ./VoicyRecorder/ -d ./VoicyRecorderTests/Helper/Mock/MockResults.swift -i VoicyRecorder --mock-final --enable-args-history
Start...
["Process input mock files..."]
["Took", 5.900859832763672e-05]
["Process source files and generate an annotated/protocol map..."]
MockoloFramework/SourceParser.swift:103: Fatal error: The operation couldn’t be completed. (SwiftSyntax.ParserError error 1.)
```

szk2s:
I have same issue with Xcode 13.3. I'm using Mockolo v1.7.0 installed by Homebrew.

psredzinski:
1.7.0 works on 13.3 when installed using MINT

wickwirew:
Ran into the same issue. Worked if I cloned it and ran it locally. Reinstalling and building from source fixed it for me.
```
brew reinstall --build-from-source mockolo
```

maxwellE:
I think this is an issue caused by brew and not the source code, can we open a ticket there and close this one?

keshavgn:
Hi 
I have version `1.7.0` and Xcode 13.4.1, getting the same error

```
Start...
["Process input mock files..."]
[...]

PR #201, opened 2022-08-25
Fixes brew command in the README

Solace-Studios:
Update README to use brew command that avoids the error in [issue 182](https://github.com/uber/mockolo/issues/182).


maxwellE:
If we are building from source should we just remove the brew guidance and advise users to clone the repo and build via SPM? It seems like brew in incompatible with this library due to the transitive dependency on swift-syntax (this is a concept that brew cannot handle)

sidepelican:
Issues around `lib_InternalSwiftSyntaxParser.dylib` are solved at version `2.0.0`.

PR #204, opened 2022-09-12
Add workaround for Xcode 14

akkyie:


akkyie:
Sorry, wrong upstream.

PR #208, opened 2022-11-21
Use SwiftSyntax.TokenKind to determine a String is SwiftKeyword or not.

sidepelican:
`extension` is keyword of Swift, but it is not registered in `String.SwiftKeywords` so it doesn't be escaped.

In the new approach, use `SwiftSyntax.TokenKind` to determine a String is SwiftKeyword or not.

### DIFF
--- a/Sources/MockoloFramework/Utils/StringExtensions.swift
+++ b/Sources/MockoloFramework/Utils/StringExtensions.swift
@@ -15,6 +15,7 @@
 //
 
 import Foundation
+import SwiftSyntax
 
 extension Int {
     var tab: String {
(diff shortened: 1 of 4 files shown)


Sources/MockoloFramework/Utils/StringExtensions.swift
```diff
@@ -251,3 +242,174 @@ extension StringProtocol {
         return self.dropFirst(String.importSpace.count).trimmingCharacters(in: .whitespaces)
     }
 }
+
+extension TokenKind {
+    init?(keyword: String) {
```
sidepelican, line 247: This initialize is very fat, but this will be implemented in future release of SwiftSyntax. https://github.com/apple/swift-syntax/blob/main/Sources/SwiftSyntax/gyb_generated/TokenKind.swift#L146

uhooi (approved):
LGTM :)

PR #213, opened 2023-02-06
Support `some` parameter's handler

### LINKED ISSUE #212: Consider method that has `some` parameter.
## Problem

When we try to create a protocol that has method like below,

```swift
public protocol Foo {}

public struct Bar: Foo {}

// NG
/// @mockable
public protocol Controller {
    func method(foo: some Foo) -> Int
}
```

Output mock fails to compile like this.

Screenshot 2023-02-04 at 22 11 17

## Reason & Workaround

The reason is `some` keyword can't be used at the parameter of closure if the closure is placed at return type.

```swift
public var methodHandler: ((some Foo) -> (Int))? // compile error happens
```

We can avoid this case by using `<F: Foo>` generics parameter instead of `some` keyword.

```swift
public protocol Foo {}

public struct Bar: Foo {}

// OK
/// @mockable
public protocol Controller {
    func method<F: Foo>(foo: F) -> Int
}
```

but I think it is valuable to investigate if it can be fixed or not.

## Sample to reproduce

- https://github.com/fummicc1/Mockolo_issue_212

fummicc1:
- close #212 

I added new property `isOpaqueType` to take care of `some` keyword placed at paramter.

## After this fix

|source| mock |
|---|---|
|Screenshot 2023-02-07 at 0 29 26|Screenshot 2023-02-07 at 0 29 20|


uhooi:
@fummicc1 thx for PR!!
I would like you to write a unit test :pray:

fummicc1:
@uhooi 

Thanks for quick reply!

I tried to add two unit tests for my fixes.

sidepelican (commented):
ORT is shorthand of “Opaque Result Type” .
Also, if it is not a return value, I would be called as “Opaque Type”.
https://github.com/apple/swift-evolution/blob/main/proposals/0341-opaque-parameters.md

fummicc1:
@sidepelican 
Thank you!

I corrected them as you mentioned.

fummicc1:
@uhooi 
Hello~
I would like you to review this PR when you have available time.

uhooi (approved):
LGTM :)

karrotpay-ray:
@uhooi 
Hi uhooi
Could you release a new version with the merged PR?

uhooi:
@fummicc1 Granted write roll.
Feel free to cut a new tag/release (see the past ones for example); make sure to bump the version first at https://github.com/uber/mockolo/blob/master/Sources/MockoloFramework/Version.swift (e.g. 1.8.0 -> 1.8.1)

PR #217, opened 2023-03-04
Support type composition (`&` keyword)

### LINKED ISSUE #170: Issue when Generating Mocks with Protocol Composition
Howdy! I am trying to generate a mock for a protocol that inherits from other protocols using protocol composition. When testing this, I tried the following:

```
/// @mockable
protocol FirstSwiftProtocol {
    var someBool: Bool { get }

    func doSomething()
}

/// @mockable
protocol SecondSwiftProtocol {
    var someInt: Int { get set }

    func doSomethingElse() -> Bool
}

/// @mockable
protocol TestProtocol: FirstSwiftProtocol & SecondSwiftProtocol {}
```

And the result of generating mocks was the following:

```
class TestProtocolMock: TestProtocol {
    init() { }
    init(someBool: Bool = false) {
        self.someBool = someBool
    }


    private(set) var someBoolSetCallCount = 0
    var someBool: Bool = false { didSet { someBoolSetCallCount += 1 } }

    private(set) var doSomethingCallCount = 0
    var doSomethingHandler: (() -> ())?
    func doSomething()  {
        doSomethingCallCount += 1
        if let doSomethingHandler = doSomethingHandler {
            doSomethingHandler()
        }
        
    }
}
```

which fails to compile as TestProtocolMock does not conform to SecondSwiftProtocol. Interestingly, if I wrap the conformance:
```
protocol TestProtocol: (FirstSwiftProtocol & SecondSwiftProtocol) {}
```

The mock is:

```
class TestProtocolMock: TestProtocol {
    init() { }


}
```
Additionally, if I use a typealias, I get the same empty init mock as when I wrap the conformance. 

Looking for any insight as to why this may be happening. Thank you!

fummicc1:
## Overview
When use `&` keyword to conform to multiple protocols, mockolo only picks up the first protocol.

This issue is already mentioned in #170 

- target mockable

```swift
protocol FirstSwiftProtocol {
    var someBool: Bool { get }

    func doSomething()
}

protocol SecondSwiftProtocol {
    var someInt: Int { get set }

    func doSomethingElse() -> Bool
}

/// @mockable
protocol TestProtocol: FirstSwiftProtocol & SecondSwiftProtocol {
    func doSomethingSpecial()
}

```

- Output mock (members in `SecondSwiftProtocol` are not implemented.)

```swift
///
/// @Generated by Mockolo
///






class TestProtocolMock: TestProtocol {
    init() { }
    init(someBool: Bool = false) {
        self.someBool = someBool
    }


    private(set) var someBoolSetCallCount = 0
    var someBool: Bool = false { didSet { someBoolSetCallCount += 1 } }

    private(set) var doSomethingCallCount = 0
    var doSomethingHandler: (() -> ())?
    func doSomething()  {
        doSomethingCallCount += 1
        if let doSomethingHandler = doSomethingHandler {
            doSomethingHandler()
        }
        
    }

    private(set) var doSomethingSpecialCallCount = 0
    var doSomethingSpecialHandler: (() -> ())?
    func doSomethingSpecial()  {
        doSomethingSpecialCallCount += 1
        if let doSomethingSpecialHandler = doSomethingSpecialHandler {
            doSomethingSpecialHandler()
        }
        
    }
}


```

- close #170

## Solution

checking if inheritedType is `CompositionType` or not.

- Abstract Syntax Tree
Screenshot 2023-03-04 at 16 35 08

### DIFF
--- a/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
+++ b/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
@@ -99,19 +99,24 @@ extension TypeInheritanceClauseSyntax {
     var types: [String] {
         var list = [String]()
         for element in self.inheritedTypeCollection {
-            if let composition = element.typeName.as(CompositionTypeSyntax.self) {
-                // Match Case: use `&` keyword to conform to multiple protocols.
-                // example: `A: B & C`
-                for compositionElement in composition.elements {
-                    if let elementName = compositionElement.firstToken?.text {
-                        list.append(elementName)
-                    }
-                }
-            } else if let elementName = element.firstToken?.text {
-                list.append(elementName)
-            }
+            let elementNameList = parseElementType(type: element.typeName)
+            list.append(contentsOf: elementNameList)
         }
         return list
+    }
+
+    private func parseElementType(type: TypeSyntax) -> [String] {
+        if let simpleTypeIdentifier = type.as(SimpleTypeIdentifierSyntax.self) {
+            // example: `protocol A: B {}`
+            return [simpleTypeIdentifier.name.text]
+        } else if let tupleType = type.as(TupleTypeSyntax.self) {
+            // example: `protocol A: (B) {}`
+            return tupleType.elements.map(\.type).map(parseElementType(type:)).flatMap { $0 }
+        } else if let compositionType = type.as(CompositionTypeSyntax.self) {
+            // example: `protocol A: B & C {}`
+            return compositionType.elements.map(\.type).map(parseElementType(type:)).flatMap { $0 }
+        }
+        return []
     }
 
     var typesDescription: String {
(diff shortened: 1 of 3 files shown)


sidepelican (approved):
LGTM👍

This will not work in nested tupled protocol but maybe enough with this implementation in majour usecase.

fummicc1:
@sidepelican 

Thank you for reviewing!

> This will not work in nested tupled protocol but maybe enough with this implementation in majour usecase.

I wasn't aware of this case, you mean `protocol A: (B & C)` was not correctly parsed, right?

Screenshot 2023-03-06 at 12 43 54


I think this problem can be fixed at https://github.com/uber/mockolo/pull/217/commits/6253922b620d2362ac0c6369b85b5fcf698c376a .
What do you think of it?

sidepelican (approved):
Right. Perfect 👍.

PR #223, opened 2023-03-13
Use lib_InternalSwiftSyntaxParser resolved by SPM

### LINKED ISSUE #219: Xcode14.2 Fatal error: The operation couldn’t be completed. (SwiftSyntaxParser.ParserError error 1.)
After updating Mockolo version 1.8.1, we are facing a folloiwing issue.

> MockoloFramework/SourceParser.swift:105: Fatal error: The operation couldn’t be completed. (SwiftSyntaxParser.ParserError error 1.)

Till Mocolo version 1.7.1 every things works fine.

Seems like it related swift-syntax version.
https://github.com/uber/mockolo/blob/2b425df144b0a1e61bf2202d481637741b2b3270/Package.swift#L18

ENVIRONMENT
- ARM64 M1

How the Mockolo is integated in project?
- Extracting the [mockolo.tar.gz](https://github.com/uber/mockolo/releases/download/1.8.1/mockolo.tar.gz)

Other Context
- We tried it with `1.8.0` it's also not working, aslo figure out that mockolo Version file is not update with which is fixed in v1.8.1
- Fetching the ` 1.8.1` via [MINT](https://github.com/yonaskolb/Mint) works

fummicc1:
## Overview

One of the install option; downloading mockolo.tar.gz does not work if  we generate mockolo binary with`lib_InternalSwiftSyntaxParser` which is placed inside Xcode toolchain.

I think there is no need to copy `lib_internalSwiftSyntaxParser.dylib` from xcode toolchain and we can just use one resolved by running `swift build`.

please feel free to correct me if there is a better way.

## Issue

- close #219

### DIFF
--- a/install-script.sh
+++ b/install-script.sh
@@ -69,7 +69,6 @@ swift build -c release --arch arm64 --arch x86_64
 cd .build/apple/Products/Release
 
 echo "** Install..."
-cp "$(xcode-select -p)"/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/lib_InternalSwiftSyntaxParser.dylib . 
 
 install_name_tool -change @rpath/lib_InternalSwiftSyntaxParser.dylib @executable_path/lib_InternalSwiftSyntaxParser.dylib "$TARGET"
 


fummicc1:
but, in the near future, swift-syntax will release `SwiftParser` which does not require `lib_InternalSwiftSyntaxParser.dylib`.
so I think we should replace `SwiftSyntaxParser` with `SwiftParser` as an other task.

install-script.sh
```diff
@@ -69,7 +69,10 @@ swift build -c release --arch arm64 --arch x86_64
 cd .build/apple/Products/Release
 
 echo "** Install..."
-cp "$(xcode-select -p)"/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/lib_InternalSwiftSyntaxParser.dylib . 
+if [[ ! -e ./lib_InternalSwiftSyntaxParser.dylib ]]; then
+    echo "** Copy lib_InternalSwiftSyntaxParser.dylib from Xcode toolchain..."
+    cp "$(xcode-select -p)"/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/lib_InternalSwiftSyntaxParser.dylib .
```
sidepelican, line 74: Is this fallback necessary?
In almost all cases, this will not work and the original issue will be hidden.
This script should fail here.

sidepelican (changes requested):

install-script.sh
fummicc1: Thank you for your review.

> Is this fallback necessary?

In my understanding, if swift version is under `5.6`, `lib_InternalSwiftSyntaxParser` will not be prepared in resolving by SPM.
so in case user's swift environment is under swift 5.6, I didn't delete option to copy from xcode toolchain.

sidepelican: In my understand, this script is used for distribution.

For legacy swift user, artifacts exist in release page. Also they can build mockolo theirself ( this is guaranteed by https://github.com/uber/mockolo/blob/master/.github/workflows/builds.yml )
Some escape hatches exist so not needs to consider backward compatibility I think.

fummicc1: Ah, you are correct. `install-script.sh` is not an install option but distribution way.
I removed the fallback.

sidepelican (approved):
Nice work!

Issue #226, opened 2023-04-04
SwiftSyntax 508.0.0


sidepelican:
SwiftSyntax 508.0.0 has been released 🎉
https://github.com/apple/swift-syntax/releases/tag/508.0.0

The new version has no dependencies to lib_InternalSwiftSyntaxParser.dylib and has good portability.
I think it would solve whole installation problems.

sidepelican:
Maybe we can remove legacy version support because mockolo already does not have dependency to local toolchain (from SwiftSyntax 0.50700.1).

fummicc1:
🎉
I think we can replace all of `SwiftSyntaxParser` with `SwiftParser` (maybe typealias is fine?)

```swift
#if swift(<5.8)
typealias SwiftParser = SwiftSyntaxParser
#endif
```

but https://github.com/apple/swift-syntax/tree/508.0.0/Sources/SwiftParser says SwiftParser is not fully compatible with SwiftSyntaxParser.

PR #230, opened 2023-04-13
Use SwiftSyntax 508.0.0

sidepelican:
- https://github.com/uber/mockolo/issues/226

Update SwiftSyntax dependency.

## Changes

- Update development swift version to 5.8
- Not depend to lib_InternalSwiftSyntaxParser.dylib on run time. Many installation problems are solved and some legacy supports can be dropped.
- Linux support.
- async initializer ( `init() async` ) support.

## Note 

I'm planning to release this update as v2.0.0. 
After v2, building mockolo on legacy Xcode will no longer be supported. 
If someone want to build with older Xcode, it is needed to refer to the v1 version.
Basically, I don't think anyone will be troubled by this. Many other tools that use SwiftSyntax are managed like this.


Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -6,27 +6,23 @@
...
-#endif
+import SwiftParser
 
-extension SyntaxParser {
-    public static func parse(_ fileData: Data, path: String) throws -> SourceFileSyntax {
-        // Avoid using `String(contentsOf:)` because it creates a wrapped NSString.
-        let source = fileData.withUnsafeBytes { buf in
-            return String(decoding: buf.bindMemory(to: UInt8.self), as: UTF8.self)
+extension Parser {
+    public static func parse(_ fileData: Data, path: String) -> SourceFileSyntax {
```
fummicc1, line 13: Sorry to pick up tiny refactoring but shall we remove `path` parameter because it might not be used anywhere?
(Or, I think we can delete this method)

fummicc1 (approved):
Thank you for this great work! LGTM from my side. 
but it might be good to wait for a while to check if anyone who is against this changes appears.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
sidepelican: OK. This function should be considered for backward compatibility because it is public, but this function may not be used from users. it can be deleted.
memo: impl commit https://github.com/uber/mockolo/commit/622227b0c5df0fcdc9913139bbdd2ef5f1632aeb

pablocornejo:
Hi, it seems I arrived too late, as this was merged before #231, where a bug is fixed. It'd be great if we can release one last  v1 version (1.8.3) that contains all possible bug fixes before releasing the aforementioned 2.0.0, which contains breaking changes for those who need to use Xcode versions earlier than 14.3. 

That way we would avoid depriving v1 users of the bug fixes that have already been implemented and would be valuable for them.

Do you think we could revert these changes, release one more fully-patched v1 version, and then re-introduce these breaking changes once the v2 release is imminent? If you have other solutions to deliver #231 and other potential bug fixes to v1 users before the v2 release please share them.

fummicc1:
thank you for your comments!

maybe you think mockolo cli users will not be able to execute v2.0.0 mockolo cli binary?
I also had had your concern as well but I have learned that this breaking changes will affect to users who build mockolo source code (so, affects only to mockolo contributors or MockoloFramework users).

so we can build v2.0.0 mockolo binary with Xcode14.2 or below, in my understanding.

pablocornejo:
Hi. Yes, I was referring to the users who have mockolo as a dependency and build it from source. If they cannot use v2 because, for example, they need to use a lower swift-tools-version in their Package.swift, they wouldn't be able to get the bug fixes like #231. 

The ideal thing would've been to merge #231 before this PR, so a 1.8.3 version can be released with bug fixes but without the breaking changes. But given that this was already merged, do you think we'd be able to do something like what I mentioned on my previous comment to deliver #231 on a 1.8.3 version before releasing v2?

sidepelican:
Releasing v1.8.3 is difficult because mockolo v1 cannot build on macOS 13 using Xcode 14.3. This means there is a pain to provide homebrew bottle.
Unfortunatlly, #231 was marged after brew supported Xcode14.3, reguradless of #230, it is difficult.

On the other hand, this PR set swift-tools-version to 5.8 but this was mistake.
To support build on macOS12 (Xcode14.0~14.2), it should be 5.7.

fummicc1:
I believe it may be difficult to support Swift 5.6, as Xcode 13.3 does not work on Ventura OS, and Xcode 14.3 is incompatible with any OS other than Ventura. Though there is a possibility that Linux users will demand compatibility with Swift 5.6, it shouldn't break the v1 mockolo. Currently, we have adequate support to build the mockolo source using Swift 5.7 or later, so we are prepared to release mockolo v2 whenever the time is right.

Issue #224, opened 2023-03-15
[bug] Duplicated underlying var declarations generated for protocol mocks with init requirements


pablocornejo:
Hi, I've found a bug where when mocking a protocol that has multiple init requirements that contain a parameter with the same name, the generated underlying var declaration gets duplicated.

For example, generating a mock for:

```swift
protocol MyProtocol {
    init(param: Any)
    init(param: Any, anotherParam: String)
}
```

Results in:

```swift
class MyProtocolMock: MyProtocol {
        private var _param: Any!
    private var _anotherParam: String!
    private var _param: Any!
    public init() { }
    required init(param: Any) {
        self._param = param
    }
    required init(param: Any, anotherParam: String = "") {
        self._param = param
        self._anotherParam = anotherParam
    }


}
```

...which fails to compile: `error: invalid redeclaration of '_param'`

A similar thing happens if using different types for param:

```swift
protocol MyProtocol {
    init(param: Int)
    init(param: Any)
}
```

Results in:

```swift
class MyProtocolMock: MyProtocol {
        private var _param: Any!
    private var _param: Int!
    public init() { }
    required init(param: Any) {
        self._param = param
    }
    required init(param: Int = 0) {
        self._param = param
    }


}
```

...`error: invalid redeclaration of '_param'`

fummicc1:
@pablocornejo 
Thank you for sharing!
If you have a solution, feel free to send a PR.😄
I will start to investigate this issue.

fummicc1:
Thank you. I could reproduce this issue.

After my investigation, I would like to confirm with you about the expected behavior.

In my opinion, the followings are expected behaviors.

#### duplicated init param name

- input

```swift
/// @mockable
protocol MyProtocol {
    init(param: Int)
    init(param: Any)
}
```

Before thinking the solution, let me say this code is not common in general. We can avoid this case by using typealias.

```swift
/// @mockable
protocol MyProtocol {
    associatedtype T
    init(param: T)
}

class MyProtocolMock: MyProtocol {
        private var _param: T!
    init() { }
    required init(param: T) {
        self._param = param
    }

    typealias T = Any
}
```

This is why we don't have to take care of this problem for now. How do you think of it?


#### duplicated variables to be used in initializer

-  input

```swift
protocol MyProtocol {
    init(param: Any)
    init(param: Any, anotherParam: String)
}
```

- output

```swift
///
/// @Generated by Mockolo
///






class MyProtocolMock: MyProtocol {
[...]

PR #231, opened 2023-04-14
Avoid to generate duplicated variable name

fummicc1:
## Overview

If mockable has multiple initializer which have same parameter name, output mock can't be built due to conflict of variable name.

This fix aims to solve #224 .

## After fix

- Input

```swift
/// @mockable
protocol MyProtocol {
    init(param: Any)
    init(param: String)
    init(param: any Sequence<Character>)
}
```

- Output

```swift
class MyProtocolMock: MyProtocol {
        private var _param: Any!
    init() { }
    required init(param: Any) {
        self._param = param
    }
    required init(param: String = "") {
        self._param = param
    }
    required init(param: any Sequence<Character>) {
        self._param = param
    }


}
```


Sources/MockoloFramework/Models/ClassModel.swift
```diff
@@ -25,7 +25,7 @@ final class ClassModel: Model {
     let identifier: String
     let declType: DeclType
     let entities: [(String, Model)]
-    let initParamCandidates: [Model]
+    let initParamCandidates: [VariableModel]
```
fummicc1, line 28: I think all of `Model` assigned to `initParamCandiates` are `VariableModel`. so type infromation should denote it.

Sources/MockoloFramework/Models/ClassModel.swift
```diff
@@ -55,7 +55,32 @@ final class ClassModel: Model {
...
+            useTemplateFunc: useTemplateFunc,
+            useMockObservable: useMockObservable,
+            allowSetCallCount: allowSetCallCount,
+            mockFinal: mockFinal,
+            enableFuncArgsHistory: enableFuncArgsHistory,
+            disableCombineDefaultValues: disableCombineDefaultValues,
+            initParamCandidates: initParamCandidates,
+            declaredInits: declaredInits,
+            entities: entities
+        )
```
fummicc1, line 84: I would like to make program more readable. if there is a better format, please let me know. :)

Sources/MockoloFramework/Templates/ClassTemplate.swift
```diff
@@ -293,3 +305,19 @@ extension ClassModel {
...
+    }
+
+    /// Check if there is same parameter name with different typeName.
+    func checkHasConflictedParam(name: String, typeName: String) -> Bool {
+        let sameParams = findParams(
+            name: name
+        )
+        return !sameParams.allSatisfy { $0.type.typeName == typeName }
+    }
+}
```
fummicc1, line 323: placing specific usecase logic at the bottom of file.

fummicc1:
To collaborators, could you review my PR when your spare time :pray:

Sources/MockoloFramework/Templates/ClassTemplate.swift
```diff
@@ -169,13 +171,23 @@ extension ClassModel {
...
+                    let shouldEraseType = declaredInitParamsPerInit
+                        .flatMap { $0 }
+                        .checkHasConflictedParam(
+                            name: p.name,
+                            typeName: p.type.typeName
+                        )
+                    return p.asInitVarDecl(eraseType: shouldEraseType)
+                }
+                return nil
             }
```
sidepelican, line 189: Simplify without extension

```suggestion
        let extraVarsToDecl = Dictionary(
            grouping: declaredInitParamsPerInit.flatMap {
                $0.filter { !extraInitParamNames.contains($0.name) }
            },
            by: \.name
        )
            .compactMap { (name: String, params: [ParamModel]) in
                let shouldErase = params.contains { params[0].type.typeName != $0.type.typeName }
                return params[0].asInitVarDecl(eraseType: shouldErase)
            }
            .sorted()
```

sidepelican (changes requested):

Sources/MockoloFramework/Models/ClassModel.swift
fummicc1: > Why you needs this change?

This is to improve the readability. `Model` is a bit unclear to imply what `initParamCandidates` means.
If we can treat `initParamCandidates` as `VariableModel` (or some specific Model Type), it might be helpful for others to understand what it is.

> Is there any reason that initParamCandidates should always be VariableModel?

In my understanding, all of candidates to use in initializer, so that mock class can capture init's parameter, are `VariableModel` and I couldn't come up with the counter-example.

[This block](https://github.com/uber/mockolo/blob/68eedb21a810986e936140ff4dc6662b14058148/Sources/MockoloFramework/Models/ParsedEntity.swift#L42-L58) returns only `VariableModel` because only `VariableModel.canBeInitParam` returns `true` and other Model returns `false`. [link](https://github.com/uber/mockolo/blob/68eedb21a810986e936140ff4dc6662b14058148/Sources/MockoloFramework/Models/Model.swift#L109-L111)

- example

```swift
/// @mockable
protocol MyProtocol {
    var param: Int { get } // this is initParamCandidate.
    init(param: Any)
    init(param: String)
    init(with param: [Character])
}
```



but this is kind of IMO. I need your opinion.🙏

Sources/MockoloFramework/Templates/ClassTemplate.swift
```diff
@@ -293,3 +305,19 @@ extension ClassModel {
         }
     }
 }
+
+fileprivate extension Array where Element == ParamModel {
```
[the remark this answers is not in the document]
fummicc1: I deleted here and use your request changes!

Sources/MockoloFramework/Models/ClassModel.swift
sidepelican: I didn't aware that `sortedInitVars` do filtering.
I also found that `initParamCandidates` is not a common one but a high context one.

Maybe there are few effects caused by changing the type of the property. OK.

sidepelican: You can also remove `var canBeInitParam: Bool { get }` requirement in `Model`.

sidepelican (approved):
LGTM. Nice work!

PR #233, opened 2023-04-28
Add support to build on macOS 12

sidepelican:
To support building homebrew bottle on macOS 12, set swift-tools-version to 5.7.

(The latest Xcode that macOS11 can use is 13.2.1, and it is too old so I give up.


fummicc1 (approved):
LGTM.

I think it might be better to update the description of [system requirements in README](https://github.com/uber/mockolo#system-requirements), at the same time, though I shoud have mentioned this in the previous PR :pray:

Issue #175, opened 2022-01-17
Mock for Combine


keithcml:
I have tried to annotate with the following for my AnyPublisher

```swift
/// @mockable(combine: publisher = CurrentValueSubject)
```
But the generator produces PassthroughSubject as

```swift
public private(set) var publisherSubject: PassthroughSubject...
```

I would expect there is a CurrentValueSubject instead.

ffittschen:
We are currently facing the same issue. After investigating it a bit, it only fails to generate the mock as a `CurrentValueSubject`, if it can't find any default value. E.g. in the following example, mockolo will generate a `fooPublisherSubject = PassthroughSubject<FooEnum, Never>()` and a `barSubject = CurrentValueSubject<Bool, Never>(false)`

```swift
public enum FooEnum {
    case one, two, three
}

/// @mockable(combine: fooPublisher = CurrentValueSubject; barPublisher = CurrentValueSubject)
public protocol SomeStream: AnyObject {
    var fooPublisher: AnyPublisher<FooEnum, Never> { get }
    var barPublisher: AnyPublisher<Bool, Never> { get }
}
```

In [these lines of code](https://github.com/uber/mockolo/blob/master/Sources/MockoloFramework/Templates/VariableTemplate.swift#L154-L158), mockolo will decide to fall back to a `PassthroughSubject` in case it can't find a default value.

As mockolo is already capable of adding parameters to the `init` method in the case of assigning an AnyPublisher to a property, we might be able to extend it to add parameters for types of the CurrentValueSubjects as well?

maxwellE:
I believe this was fixed in #186 ?

uhooi:
@keithcml Can you confirm that it is fixed?

kernandreas:
Just ran in the same situation. Thank you @ffittschen for the investigation. How did you work around in your code base?

Btw. #186 does not fix the root cause. It adds a new flag to disable Subject synthesis for all AnyPublisher properties.

fummicc1:
I try to fix this issue by adding a new metadata to set a default value of CurrentValueSubject  in https://github.com/uber/mockolo/pull/234.

If you have any advice, please let me know.

fummicc1:
After reviewing my idea, I believe we can resolve this issue by using the `--disable-combine-default-values` option. Therefore, I'll be closing this PR. Please feel free to reopen it whenever you wish to continue discussing this topic.

PR #234, opened 2023-05-03
[Combine] can set default value to `CurrentValueSubject`

fummicc1:
## Overview

I added new metadata to set an explicit default value for CurrentValueSubject in a Mock object, which is used to manage the Publisher stream.

### Problem

Currently, if publish type ( type `A` of `Publisher<A, B>`) is not kind of primitive type, subject becomes `PassthroughSubject` even if you mark it as `CurrentValueSubject` by `@mockable(combine: aPublisher = CurrentValueSubject)`.

### Note
This PR only addresses Combine, but RxSwift also has the same problem. I believe this should be considered a separate issue because the implementation differs from the Combine one.

### Example

Screenshot 2023-05-03 at 23 49 51

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,6 +361,28 @@ public class FooMock: Foo {
 }
 ```
 
+Warning: If you would like to send a custom data event via `Publisher`, it is hard to specify default value from mockolo this is why `PassthroughSubject` is used insteaad of `CurrentValueSubject` regardless of your specification.
+As a solution, you can set a default value via metadata like the following.
+
+```swift
+/// @mockable(combine: fooPublisher = CurrentValueSubject; default = CustomType())
+public protocol Foo {
+    var fooPublisher: AnyPublisher<CustomType, CustomError> { get }
+}
+```
+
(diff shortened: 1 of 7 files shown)


fummicc1:
@sidepelican 

I'm in no hurry but could you review this PR when you have time :pray:

sidepelican:
This change looks good, but I don't want to expand the Combine related functions.
In case of #175, it can also be avoided by using the `--disable-combine-default-values` option. This change doesn't seem necessary.

If we are going to be hospitable to #175, I think it would be to display warning messages when the CurrentValueSubject fall through implicitly to the PassthroughSubject.
On that time, it would also be good to display the existence of the `--disable-combine-default-values` option.

fummicc1:
Yes. honestly, I did not know that flag. It is what I would like! Thank you.

>  it would be to display warning messages when the CurrentValueSubject fall through implicitly to the PassthroughSubject.

I will try to address to show the warning message instead of this PR. 👍

PR #235, opened 2023-05-11
Wrap underlyingType if the type has existential any

sidepelican:
Add support for existential `any`.

## Current behavior

```swift
protocol P {
    var foo: any P { get }
}
```

↓

```swift
class PMock: P {
...
    private(set) var fooSetCallCount = 0
    private var _foo: any P!  { didSet { fooSetCallCount += 1 } }
    var foo: any P {
        get { return _foo }
        set { _foo = newValue }
    }
}
```

This is wrong. `any P!` must be `(any P)!`.

## Expected

```swift
    private(set) var fooSetCallCount = 0
    private var _foo: (any P)!  { didSet { fooSetCallCount += 1 } }
    var foo: any P {
        get { return _foo }
        set { _foo = newValue }
    }
```

## Misc

Keep current behavior for non `any` marked protocol for legacy codes.


Tests/TestExistentialAny/FixtureExistentialAny.swift
```diff
@@ -0,0 +1,52 @@
+let existentialAny = """
+/// \(String.mockAnnotation)
+protocol ExistentialAny {
+    var foo: P { get }
+    var bar: any P { get }
+    var baz: any P & Q { get }
+    var qux: (any P) -> any P & Q { get }
```
fummicc1, line 7: Is it possible to check if `primary associated type` is also working with this change?

```
var a: any Collection<P> { get }
```

fummicc1 (approved):
Thank you so much! 🚀

PR #243, opened 2023-06-20
Only consider relevant files when collecting import statements.

### LINKED ISSUE #242: import collection is overly broad
Today, mockolo is collecting import statements from all the files that it sees.  This includes the files sent in via the `--mocks` also.  We should be a lot more selective. Only files that have relevant protocols should be considered when gathering imports.

rudro:
- There's no need to include all the imports we see.
- Fixes #242

### DIFF
--- a/Sources/MockoloFramework/Operations/ImportsHandler.swift
+++ b/Sources/MockoloFramework/Operations/ImportsHandler.swift
@@ -11,7 +11,8 @@ func handleImports(pathToImportsMap: ImportMap,
                    pathToContentMap: [(String, Data, Int64)],
                    customImports: [String]?,
                    excludeImports: [String]?,
-                   testableImports: [String]?) -> String {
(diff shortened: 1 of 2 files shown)


Sources/MockoloFramework/Operations/ImportsHandler.swift
```diff
@@ -11,15 +11,17 @@ func handleImports(pathToImportsMap: ImportMap,
                    pathToContentMap: [(String, Data, Int64)],
                    customImports: [String]?,
                    excludeImports: [String]?,
-                   testableImports: [String]?) -> String {
+                   testableImports: [String]?,
+                   relevantPaths: [String]) -> String {
```
ryanaveo, line 15: nit: we should document this parameter's purpose

ryanaveo (approved):

PR #255, opened 2024-04-17
Update swift-syntax 510

sidepelican:
Update swift-syntax.
No codebase changes were necessary.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,7 +11,7 @@ let package = Package(
         .library(name: "MockoloFramework", targets: ["MockoloFramework"]),
         ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-syntax.git", exact: "509.0.0"),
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -4,14 +4,14 @@ import PackageDescription
...
+        .macOS(.v12),
     ],
     products: [
         .executable(name: "mockolo", targets: ["Mockolo"]),
         .library(name: "MockoloFramework", targets: ["MockoloFramework"]),
-        ],
+    ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-syntax.git", exact: "509.0.0"),
+        .package(url: "https://github.com/apple/swift-syntax.git", from: "510.0.1"),
```
bielikb, line 14: Hey there, 
I think it's better to lock to a specific version of swift-syntax to avoid issues with transient dependencies.

sidepelican: That's an old style. Please take a look at the latest swift-syntax.

PR #259, opened 2024-04-26
Compilation fix for a mock where the return type of a closure is an opaque type

omarzl:
The following code:

```
protocol ProtocolReturningOpaqueTypeInClosureProtocol {
    func register(router: @autoclosure @escaping () -> (some Error))
}
```

produces this mock:
```
class ProtocolReturningOpaqueTypeInClosureProtocolMock: ProtocolReturningOpaqueTypeInClosureProtocol {
    init() { }
    private(set) var registerCallCount = 0
    var registerHandler: ((@autoclosure @escaping () -> (some Error)) -> ())?
    func register(router: @autoclosure @escaping () -> (some Error))  {
        registerCallCount += 1
        if let registerHandler = registerHandler {
            registerHandler(router())
        }
    }
}
```

causing the compilation error:
`'some' cannot appear in parameter position in result type '((@autoclosure @escaping () -> (some Error)) -> ())?'`

This pr fixes the variable `registerHandler` by generating the following mock:
```
class ProtocolReturningOpaqueTypeInClosureProtocolMock: ProtocolReturningOpaqueTypeInClosureProtocol {
    init() { }
    private(set) var registerCallCount = 0
    var registerHandler: ((@autoclosure @escaping () -> (any Error)) -> ())?
    func register(router: @autoclosure @escaping () -> (some Error))  {
        registerCallCount += 1
        if let registerHandler = registerHandler {
            registerHandler(router())
        }
    }
}
```

### DIFF
--- a/Sources/MockoloFramework/Utils/TypeParser.swift
+++ b/Sources/MockoloFramework/Utils/TypeParser.swift
@@ -600,6 +600,9 @@ public final class `Type` {
         var ret = typeName
         if let closureRng = closureRng {
             let left = ret[ret.startIndex..<closureRng.lowerBound]
+            let right = ret[closureRng.lowerBound..<ret.endIndex]
+            ret = left + right.replacingOccurrences(of: String.some, with: String.any)
+            
             for item in typeParamList {
                 if isEscaping, left.literalComponents.contains(item) {
                     ret = String.anyType
--- a/Sources/MockoloFramework/Utils/TypeParser.swift
+++ b/Sources/MockoloFramework/Utils/TypeParser.swift
@@ -601,7 +601,7 @@ public final class `Type` {
         if let closureRng = closureRng {
             let left = ret[ret.startIndex..<closureRng.lowerBound]
             let right = ret[closureRng.lowerBound..<ret.endIndex]
-            ret = left + right.replacingOccurrences(of: String.some, with: String.any)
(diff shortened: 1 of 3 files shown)


sidepelican (commented):
I wonder if this implementation would work well with the following code?🤔

```swift
protocol P {
     func f(_: () -> (MyAwesomeType))
}
```

omarzl:
> I wonder if this implementation would work well with the following code?🤔


Hello, I am only replacing the word `some` to `any` when the return type is `some MyType`

These other examples work well:

```
protocol Foo {
    func register(router: () -> (Bar))
    func register<T>(router: () -> (T))
}

protocol Bar {}
```
The generated mocks are:
```
var registerRouterHandler: ((() -> (Bar)) -> ())?
var registerRouterTHandler: ((() -> (Any)) -> ())?
```

sidepelican (changes requested):
I've confirmed that `MyAwesomeType` changes to `MyAweanyType`.
It seems sufficient to just change the replacement logic to `s/some /any /g`.

omarzl:
oh right, sorry I misunderstood you
I updated the fix and the tests to consider this scenario, thanks!

sidepelican (approved):
Maybe work with most codes. Thanks!

sidepelican:
Sorry I forgot launch workflow.

Issue #249, opened 2023-11-21
Async functions are unsafe in generated mocks, and can crash if used from multiple threads simultaneously


stuaustin:
## What did you do?

Created a mock for a protocol with an async function, e.g.
```
/// @mockable
protocol APIClient {
    func fetchData(_ id: AnyObject) async -> Int
}
```

This is the generated mock for the example code above:
```
class APIClientMock: APIClient {
    init() { }


    private(set) var fetchDataCallCount = 0
    var fetchDataArgValues = [AnyObject]()
    var fetchDataHandler: ((AnyObject) async -> (Int))?
    func fetchData(_ id: AnyObject) async -> Int {
        fetchDataCallCount += 1
        fetchDataArgValues.append(id)
        if let fetchDataHandler = fetchDataHandler {
            return await fetchDataHandler(id)
        }
        return 0
    }
}
```

## What did you expect to happen?

The mock should be thread safe, or at least there should be a way to generate the mock in a thread safe manor.
Some possibilities:
- Isolate the mocked function to an actor, e.g. MainActor. That way the mock's logic always runs on the same thread.
- Isolate the history properties to an actor.
- Isolate the entire class to an actor.
- Use an actor instead of a class.

## What happened instead?

The mock is not thread safe.

If you run the following code, the mock will crash with an EXC_BAD_ACCESS error on the `fetchDataArgValues.append(id)` line above.

```
        let mockAPI = APIClientMock()
        await withTaskGroup(of: Int.self, body: { taskGroup in
            for _ in 0..<1_000 {
                taskGroup.addTask {
                    await mockAPI.fetchData(NSObject())
                }
            }
            await taskGroup.waitForAll()
        })
```

This is because the `fetchData` function and `fetchDataArgValues` property are not actor isolated, and since the `fetchData` function is async it can be run on any thread.
When multiple threads simultaneously try to read/write `fetchDataArgValues` this is unsafe, and the resultant EXC_BAD_ACCESS is caused by this.

## Mockolo Environment

**Mockolo Version:** 2.0.1
**Dependency Manager:** mint
**Xcode Version:** 15.0
**Swift Version:**  Swift 5.9 (5.9.0.128.108)

bielikb:
Hey Uber team,

our unit test suite faces reliability issues due to this issue. 

What would be the best way to move this issue forward?

Best,
Boris

bielikb:
[...]

PR #260, opened 2024-05-21
Added override for specifying the global actor of mocked async functions

stuaustin:
# Issue
#249  Async functions are unsafe in generated mocks, and can crash if used from multiple threads simultaneously

# Description
This change lets you specify the global actor that is used for your generated mock's async functions. This can be used to solve the problem described in #249 as when used there is no longer the chance for the async functions to be executed multiple times simultaneously.
I have allowed the user to specify the global actor so that they can choose to use a custom GlobalActor instead of assuming they always want to use the MainActor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,32 @@ This will generate:
 public class FooMock: FooProtocol { ... }
 ```
 
+To override the actor used for a mock's async functions:
+```swift
+/// @mockable(override: asyncFunctionGlobalActor = MainActor)
+public protocol FooProtocol {
+    func asyncFunction() async -> Bool
+}
+```
+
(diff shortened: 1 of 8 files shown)


sidepelican:
Thank you for your contribution.

While this feature is opt-in, I don't believe using a GlobalActor is the fundamental solution. Introducing locks would be the correct approach to address exclusive access issues.

However, mockolo currently lacks a suitable lock implementation. `NSLock` depends on Foundation, and swift-atomics might not always be available. I’m looking forward to the standard library introducing a `Mutex`.
https://github.com/apple/swift-evolution/blob/main/proposals/0433-mutex.md.

Issue #261, opened 2024-06-13
Support for Typed throws


sutakahashi1:
Please add support for Typed throws as described in [SE-0413](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0413-typed-throws.md).

For example:

```swift
func throwsAnything() throws(any Error) { ... }
```

Thank you.

fummicc1:
This feature is introduced in #267 and #262 . Please try it.

wvteijlingen-npo:
Thanks for introducing  this feature! Would it be possible to make a new official release containing this change? Right now we're blocked from using mockolo from homebrew, because that version doesn't support typed throws yet.

PR #262, opened 2024-06-30
Correct the `throws` clause in the closure within template logic to support TypedThrow.

fummicc1:
In this PR, I fixed tiny bug to support typed throw for closure handler. Also, added some changes as nits.

- https://github.com/uber/mockolo/issues/261


Tests/TestFuncs/TestFuncThrow/FixtureFuncThrow.swift
```diff
@@ -1,23 +1,15 @@
...
-let funcThrow2 = """
-import Foundation
-
-/// \(String.mockAnnotation)
-protocol FuncThrow {
-func update<T>(arg1: T, arg2: () throws -> T) rethrows -> T
-func update<T, U>(arg1: T, arg2: @escaping (U) throws -> ()) throws -> ((T) -> (U))
-}
-"""
-
```
fummicc1, line 13: FYI

This test-case is not used and is covered by `funcThrow` test-case. So I deleted it.

Sources/MockoloFramework/Utils/TypeParser.swift
```diff
@@ -510,6 +510,25 @@ public final class `Type` {
...
+                guard let prefixRange = clause.range(of: "\(String.throws)(") else {
+                    return nil
+                }
+                let endIndex = clause.dropLast().endIndex
+                return String(
+                    clause[prefixRange.upperBound..<endIndex]
+                )
+            }
+            .first
+    }
```
fummicc1, line 531: I am not sure if splitting by `.whitespace` is suitable or not. but this code works in added test-cases.

As far as I investigated, 
- `any Error` can't be thrown type because it's not a concrete type. 
- `some Error` can't come here due to an issue related to performance (https://github.com/swiftlang/swift-evolution/blob/main/proposals/0413-typed-throws.md#opaque-thrown-error-types).

Package.swift
```diff
@@ -11,7 +11,7 @@ let package = Package(
         .library(name: "MockoloFramework", targets: ["MockoloFramework"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-syntax.git", from: "510.0.1"),
+        .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "600.0.0-prerelease-2024-06-12"),
```
fummicc1, line 14: to use `ThrowsClauseSyntax`, I updated swift-syntax.

Sources/MockoloFramework/Utils/StringExtensions.swift
```diff
@@ -108,10 +108,10 @@ extension String {
     ///
     """
 
-
     var hasThrowsOrRethrows: Bool {
         return components(separatedBy: .whitespaces).contains { component in
-            return component == .throws || component == .rethrows
+            let hasTypedThrow = hasPrefix(String.throws.withLeftParen) && hasSuffix(")")
```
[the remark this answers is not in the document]
sidepelican: It may be the time to extract this from the `FunctionEffectSpecifiersSyntax` instead of the `String`.

Sources/MockoloFramework/Utils/TypeParser.swift
sidepelican: `throws(any Error)` is entirely possible.

The Swift's syntax has become more than complicated enough to look these up from `String` alone.
More structured information should be received from the arguments.

Tests/TestFuncs/TestFuncThrow/FixtureFuncThrow.swift
```diff
@@ -1,23 +1,15 @@
...
 let funcThrow = """
 import Foundation
 
 /// \(String.mockAnnotation)
 protocol FuncThrow {
     func f1(arg: Int) throws -> String
     func f2(arg: Int) throws
+    func f3(arg: Int) throws(SomeError)
+    func f4(arg: Int) throws(SomeError) -> String
```
sidepelican, line 12: Could you add other possible patterns?

- `f() throws (MyError)`
- `f() async throws(any Error)`

sidepelican (changes requested):

Sources/MockoloFramework/Utils/TypeParser.swift
fummicc1: Thank you for your information.

> throws(any Error) is entirely possible.

I did not know `any Error` can become throws type. I need to modify the implementation.

> The Swift's syntax has become more than complicated enough to look these up from String alone.
More structured information should be received from the arguments.

Absolutely. I felt like `suffix` variable in `MethodModel` should not be `String` but more structured type.

Sources/MockoloFramework/Utils/StringExtensions.swift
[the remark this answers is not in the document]
fummicc1: @sidepelican 

I added new structure `FunctionSuffixClause` to represent more rich information in suffix.

- [909d26e](https://github.com/uber/mockolo/pull/262/commits/909d26ecaf2585b3319a5841e7b0a5591d3a56e9)

Could you give me any feedback for this approach?

Tests/TestFuncs/TestFuncThrow/FixtureFuncThrow.swift
fummicc1: Thank you!
I added these cases.: [75a0f36](https://github.com/uber/mockolo/pull/262/commits/75a0f361af0e575ae4fb2a59bb99031874a286c6)

Sources/MockoloFramework/Templates/FunctionSuffixTemplate.swift
fummicc1, line 1: review-note 

This kind of template that is not for Model but Clause is new approach.
I am not confident if it is okay or not in term of architecture.

Sources/MockoloFramework/Templates/FunctionSuffixTemplate.swift
```diff
@@ -0,0 +1,40 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+extension FunctionSuffixClause {
+    func applyFunctionSuffixTemplate(
+        forClosureTemplate: Bool
```
fummicc1, line 19: review-note 

For closure, `rethrows` has to be replaced with `throws`.

example: 

```swift
var barHandler: () throws -> Void
func foo(bar: () throws -> Void) rethrows -> Void {}

File: Sources/MockoloFramework/Utils/StringExtensions.swift (hunk 4 above)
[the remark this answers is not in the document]
Comment (fummicc1) [reply Sources/MockoloFramework/Utils/StringExtensions.swift]: Fundamental implementation for typed throw is implemented in #267 . my original implementation is unnecessary.

File: Package.resolved [hunk 7]
```diff
@@ -12,7 +12,7 @@
     {
       "identity" : "swift-syntax",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-syntax.git",
```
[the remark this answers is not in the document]
sidepelican: The original repository has been transferred into `swiftlang` so this change is right. It seems GitHub's redirection is not working in your environment but its is not problem.

Sources/MockoloFramework/Templates/ThrowingKindTemplate.swift
```diff
@@ -15,13 +15,26 @@
...
+    /// if this method is called to render handler for closure, rethrows should be replaced with throws.
+    ///
+    /// - example:
+    ///     ```
+    ///     // handler shouldn't use rethrows, instead use throws
+    ///     var fooHandler: (() -> throws -> Void) throws -> Void
+    ///     func foo(bar: () throws -> Void) rethrows -> Void
+    ///     ```
+    func applyThrowingTemplate(
+        appliesforClosureHandler: Bool
```
sidepelican, line 27: I would like to avoid, as much as possible, structuring code where special cases branch out within a common process.
This would require everyone using this feature to understand the impact caused by the flag, and the internal code could become unnecessarily complex.
I believe you can see why this is complicated, given that this pattern is widespread in the existing code.

If you only want to change `rethrows` to `throws`, I think it is enough to simply coerce it to `.any` at the caller's site.

```swift
extension ThrowingKind {
    var coerceRethrowsToThrows: ThrowingKind {
        if case .rethrows = self {
            return .throws // renamed .any to .throws
        }
        return self
    }
}
```

Sources/MockoloFramework/Templates/ThrowingKindTemplate.swift
```diff
@@ -15,13 +15,26 @@
...
+    ) -> String {
         switch self {
         case .none:
-            return nil
```
sidepelican, line 21: This change may cause effects to trivia. Sadly, it is not covered in tests.

Sources/MockoloFramework/Templates/ThrowingKindTemplate.swift
fummicc1: Got it. Thank you so much! I agree with you.

I applied it here: https://github.com/uber/mockolo/pull/262/commits/f1121102273819a4ff11f3c45142be10cde6abaf

Sources/MockoloFramework/Templates/ThrowingKindTemplate.swift
fummicc1: Thank you! I am not sure the concrete test case but fixed in 733771e .
I also guess it depends on how caller of this method process this return value (I don't understand failed case well).

sidepelican: Yes, it's a problem of the use site. You can fix either one, but there's no need to change.

Sources/MockoloFramework/Utils/TypeParser.swift
```diff
@@ -566,7 +566,7 @@ public final class SwiftType {
 
         let suffixStr = [
             isAsync ? String.async + " " : nil,
-            throwing.hasError ? String.throws + " " : nil,
+            throwing != .none ? throwing.applyThrowingTemplate()! + " " : nil,
         ].compactMap { $0 }.joined()
```
sidepelican, line 571: It want to avoid relying on internal implementations.

```swift
            let suffixStr = [
                isAsync ? String.async : nil,
                throwing.applyThrowingTemplate(),
            ].compactMap { $0 }.joined(separator: " ") + " "
```

(This snippet exists in the other 2 places. Maybe it's a time to add utility)

fummicc1: Sounds nice!  I fixed it in these 2 commits.

- [be998e1](https://github.com/uber/mockolo/pull/262/commits/be998e110e8c09a71d1917b5a01102675db78f74)
- [4a3a54a](https://github.com/uber/mockolo/pull/262/commits/4a3a54a530f677409ac34d1d2e2e19006324f35b)

sidepelican (approved):
Looks good. Nice work! thank you.

PR #265, opened 2024-10-23
Speed up aarch64 builds using Static Linux SDK

sidepelican:
## Motivation

Currently, the mockolo binary for aarch64 on Ubuntu is built on an x86_64 machine using Docker and qemu, which is extremely slow.
With the release of [Static Linux SDK](https://www.swift.org/documentation/articles/static-linux-getting-started.html) in Swift 6, we can speed up the build process for aarch64 Linux.

## Changes

In the GitHub Actions for the release build, we install the Static Linux SDK and use it. 
As a result, the size of the artifact bundle has increased. To address this, we provide a lightweight artifact bundle without Linux support for users who only use mockolo on macOS.

## Misc

Since the Static Linux SDK uses musl, mockolo will be able to run on more Linux distributions. However, we do not have a strong motivation to actively support these additional distributions, so this is more of a bonus.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,7 +54,6 @@ jobs:
           - { name: "linux-x86_64", os: ubuntu-20.04 }
           - { name: "macos-universal", os: macos-14 }
           - { name: "macos-universal", os: macos-13 }
-          - { name: "macos-universal", os: macos-12 }
     steps:
     - uses: actions/download-artifact@v4
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,8 +38,8 @@ jobs:
     - name: Upload the binary
       uses: actions/upload-artifact@v4
       with:
+        name: mockolo.${{ matrix.destination.name }}.tar.gz
         path: mockolo.${{ matrix.destination.name }}.tar.gz
-        name: mockolo.${{ matrix.destination.name }}
 
   check-portability:
     needs: build
@@ -95,6 +95,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       checksum: ${{ steps.checksum.outputs.checksum }}
+      checksum-macos: ${{ steps.checksum-macos.outputs.checksum }}
     steps:
       - uses: actions/checkout@v4
       - name: Download all artifacts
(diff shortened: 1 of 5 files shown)


sidepelican:
Test run: https://github.com/sidepelican/mockolo/actions/runs/11473420187
Result release page: https://github.com/sidepelican/mockolo/releases/releasetest-7

.github/workflows/release.yml
```diff
@@ -15,76 +17,63 @@ jobs:
...
       matrix:
         destination:
-          - { name: "ubuntu-x86_64", os: ubuntu-20.04 }
-          - { name: "macos-universal", os: macos-14 }
-    container: ${{ matrix.destination.container }}
+          - { name: "linux-aarch64", os: ubuntu-latest }
+          - { name: "linux-x86_64", os: ubuntu-latest }
+          - { name: "macos-universal", os: macos-latest }
     steps:
-    - uses: swift-actions/setup-swift@v2
```
sidepelican, line 22: `setup-swift` does not support Swift6🥲

.github/workflows/release.yml
```diff
@@ -15,76 +17,63 @@ jobs:
...
+    - if: startsWith(matrix.destination.name, 'linux')
+      uses: vapor/swiftly-action@v0.1
       with:
-        swift-version: ${{ env.SWIFT_VERSION }}
+        toolchain: ${{ env.SWIFT_VERSION }}
+    - if: startsWith(matrix.destination.name, 'macos')
+      run: sudo xcode-select -s /Applications/Xcode_16.app
```
sidepelican, line 29: swiftly's macOS support is not released yet so do the alternative here.

.github/workflows/release.yml
```diff
@@ -15,76 +17,63 @@ jobs:
...
         destination:
-          - { name: "ubuntu-x86_64", os: ubuntu-20.04 }
-          - { name: "macos-universal", os: macos-14 }
-    container: ${{ matrix.destination.container }}
+          - { name: "linux-aarch64", os: ubuntu-latest }
+          - { name: "linux-x86_64", os: ubuntu-latest }
+          - { name: "macos-universal", os: macos-latest }
     steps:
-    - uses: swift-actions/setup-swift@v2
+    - if: startsWith(matrix.destination.name, 'linux')
```
uhooi, line 24: [nits]
step order
:small_red_triangle: linux → macos → linux
:white_check_mark: linux → linux → macos or macos → linux → linux

uhooi (approved):
LGTM!

Issue #210, opened 2022-12-16
support for async properties?


alessioborraccino:
After swift 5.6 it's possible to declare in protocols async and/or throws properties:

```
/// @mockable
protocol AsyncProperties {
  var isAvailable: Bool { get async throws }
}
```

but mockolo still doesnt support this syntax, and generates just a simple `Bool` property. 

```
class AsyncPropertiesMock: AsyncProperties {
    init() { }
    init(isAvailable: Bool = false) {
        self.isAvailable = isAvailable
    }


    var isAvailableSetCallCount = 0
    var isAvailable: Bool = false { didSet { isAvailableSetCallCount += 1 } }
}
```

is there any plan to update it so that `/// @mockable` can successfully also generate an `async throws` handler for this case?

sidepelican:
It should be fixed, but I think it is a less common requirement, so it is not a priority.

alessioborraccino:
Not fixed, without a handler i cant test any throwing because the mock doesnt actually throws. Hoping in an update eventually :)

sidepelican:
shipped in 2.2.0

PR #267, opened 2024-10-25
Add async and throwing property support

sidepelican:
- https://github.com/uber/mockolo/issues/210

## Main feature

Supported `{ get async throws }`.

```swift
/// @mockable
protocol AsyncThrowsVars {
    var getAndThrows: MyValue { get throws }
    var getAndAsync: MyValue { get async }
    var getAndAsyncAndThrows: Int { get async throws(any Error) }
}
```

- Generated

```swift
class AsyncThrowsVarsMock: AsyncThrowsVars {
    init() { }
    init(getAndThrows: MyValue, getAndAsync: MyValue, getAndAsyncAndThrows: Int = 0) {
        self.getAndThrowsHandler = { getAndThrows }
        self.getAndAsyncHandler = { getAndAsync }
        self.getAndAsyncAndThrowsHandler = { getAndAsyncAndThrows }
    }


    var getAndThrowsHandler: (() throws -> MyValue)?
    var getAndThrows: MyValue {
        get throws {
            if let getAndThrowsHandler = getAndThrowsHandler {
                return try getAndThrowsHandler()
            }
            fatalError("getAndThrowsHandler returns can't have a default value thus its handler must be set")
        }
    }

    var getAndAsyncHandler: (() async -> MyValue)?
    var getAndAsync: MyValue {
        get async {
            if let getAndAsyncHandler = getAndAsyncHandler {
                return await getAndAsyncHandler()
            }
            fatalError("getAndAsyncHandler returns can't have a default value thus its handler must be set")
        }
    }

    var getAndAsyncAndThrowsHandler: (() async throws(any Error) -> Int)?
    var getAndAsyncAndThrows: Int {
        get async throws(any Error) {
            if let getAndAsyncAndThrowsHandler = getAndAsyncAndThrowsHandler {
                return try await getAndAsyncAndThrowsHandler()
            }
            return 0
        }
    }
}
```

## Side effects

- For `var foo: Int { get }`, since the value is never set through original protocol, the call count has been removed.

### DIFF
--- a/Sources/MockoloFramework/Models/ThrowingKind.swift
+++ b/Sources/MockoloFramework/Models/ThrowingKind.swift
@@ -1,0 +1,32 @@
+enum ThrowingKind: Equatable {
+    case none
+    case any
+    case `rethrows`
+    case typed(errorType: String)
+
+    var hasError: Bool {
+        switch self {
+        case .none:
+            return false
+        case .any:
+            return true
+        case .rethrows:
+            return true
+        case .typed(let errorType):
+            return errorType != .neverType && errorType != "Swift.\(String.neverType)"
+        }
+    }
+
+    var syntax: String? {
+        switch self {
+        case .none:
+            return nil
+        case .any:
+            return .throws
+        case .rethrows:
+            return .rethrows
+        case .typed(let errorType):
+            return "\(String.throws)(\(errorType))"
+        }
+    }
+}
--- a/Sources/MockoloFramework/Models/VariableModel.swift
+++ b/Sources/MockoloFramework/Models/VariableModel.swift
@@ -36,7 +36,7 @@ final class VariableModel: Model {
     }
 
     var underlyingName: String {
-        if isStatic || type.defaultVal() == nil {
+        if type.defaultVal() == nil {
             return "_\(name)"
         }
         return name
--- a/Sources/MockoloFramework/Models/VariableModel.swift
+++ b/Sources/MockoloFramework/Models/VariableModel.swift
@@ -1,6 +1,11 @@
 import Foundation
 
 final class VariableModel: Model {
+    enum GetterEffect: Hashable {
+        case async
+        case `throws`(errorType: String?)
+    }
+
     var name: String
     var type: SwiftType
     var offset: Int64
@@ -13,6 +18,8 @@ final class VariableModel: Model {
     var filePath: String = ""
     var isStatic = false
     var shouldOverride = false
+    let getterEffects: Set<GetterEffect>
+    let hasSetter: Bool
     var rxTypes: [String: String]?
     var customModifiers: [String: Modifier]?
     var modelDescription: String? = nil
@@ -40,6 +47,8 @@ final class VariableModel: Model {
          acl: String?,
          encloserType: DeclType,
          isStatic: Bool,
+         getterEffects: Set<GetterEffect>,
+         hasSetter: Bool,
          canBeInitParam: Bool,
          offset: Int64,
          rxTypes: [String: String]?,
@@ -51,6 +60,8 @@ final class VariableModel: Model {
         self.type = SwiftType(typeName.trimmingCharacters(in: .whitespaces))
         self.offset = offset
         self.isStatic = isStatic
+        self.getterEffects = getterEffects
+        self.hasSetter = hasSetter
         self.shouldOverride = encloserType == .classType
         self.canBeInitParam = canBeInitParam
         self.processed = processed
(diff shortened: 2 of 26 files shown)


sidepelican:
- [ ] it should be tested for generic(using associated type) property

Sources/MockoloFramework/Models/ThrowingKind.swift
```diff
@@ -0,0 +1,32 @@
...
+    var hasError: Bool {
+        switch self {
+        case .none:
+            return false
+        case .any:
+            return true
+        case .rethrows:
+            return true
+        case .typed(let errorType):
+            return errorType != .neverType && errorType != "Swift.\(String.neverType)"
```
sidepelican, line 16: `throws(Never)` means no throws. Without this consideration, there would be warnings in the generated code.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -368,11 +368,48 @@ extension VariableDeclSyntax {
                 typeName = vtype
             }
 
+            let storageType: VariableModel.MockStorageType
+            switch v.accessorBlock?.accessors {
```
sidepelican, line 372: This block is very complicated 🤪 but mockolo handles both class and protocol here, so it has to be this way.

Sources/MockoloFramework/Models/VariableModel.swift
```diff
@@ -29,7 +41,7 @@ final class VariableModel: Model {
     }
 
     var underlyingName: String {
-        if isStatic || type.defaultVal() == nil {
```
sidepelican, line 32: The computed property is generated as below, but I think it is meaningless, so I will remove it.
https://github.com/uber/mockolo/pull/267/files#diff-0680dee8c19da68e406e1ecf918c1c39fe5c55214b3e399ac8dd929fe6b72c7cL43-L48

Sources/MockoloFramework/Models/ThrowingKind.swift
```diff
@@ -0,0 +1,32 @@
...
+        case .none:
+            return nil
+        case .any:
+            return .throws
+        case .rethrows:
+            return .rethrows
+        case .typed(let errorType):
+            return "\(String.throws)(\(errorType))"
+        }
+    }
```
fummicc1, line 31: In my opinion, this variable can move to Templates directory by creating `ThrowingKindTemplate.swift` because mockolo clearly separates object representation and rendering  into Models and Templates.

Tests/TestFuncs/TestBasicFuncs/FixtureNonSimpleFuncs.swift
```diff
@@ -370,7 +370,7 @@ import Foundation
 
 /// \(String.mockAnnotation)
 protocol NonSimpleFuncs {
-func pass<T>(handler: @autoclosure () -> Int) rethrows -> T
+func pass<T>(handler: @autoclosure () -> Int) throws -> T
```
fummicc1, line 373: This is bug fix, right?

fummicc1 (commented):
Thank you for your PR! I will take more time to finish reviewing this PR.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -368,11 +368,48 @@ extension VariableDeclSyntax {
...
+                        if getterDecl.effectSpecifiers?.asyncSpecifier != nil {
+                            getterEffects.isAsync = true
+                        }
+                        if let `throws` = getterDecl.effectSpecifiers?.throwsClause {
+                            if let throwsType = `throws`.type {
+                                getterEffects.throwing = .typed(errorType: throwsType.trimmedDescription)
+                            } else {
+                                getterEffects.throwing = .any
+                            }
+                        }
```
fummicc1, line 388:
```suggestion
                        if let `throws` = getterDecl.effectSpecifiers?.throwsClause {
                            getterEffects.throwing = .init(`throws`)
                        }
```

this code might be duplicated.

Tests/TestFuncs/TestBasicFuncs/FixtureNonSimpleFuncs.swift
sidepelican: Sure. The original fixture code is incorrect.

Sources/MockoloFramework/Models/ThrowingKind.swift
sidepelican: That's a good insights. The modeling and code output have been described separately.
https://github.com/uber/mockolo/pull/267/commits/d958837be0888d8285587a42b78149280d2e1e1e

fummicc1 (approved):
Thank you so much!

Issue #245, opened 2023-08-15
Mock of class with computed properties causes compile error.


kaniza:
Generating a mock of a class with the following properties will result in a compile-time error in the callCount section.

var isOpen:Bool {
    return true
}

Error:. 
Cannot observe read-only property 'isOpen'; it can't change

The relevant code is this part:
override var isOpen: Bool { didSet { isOpenSetCallCount += 1 }

I think callCount is not necessary for read-only Computed Property.

mockolo: 2.0.0 (latest in Homebrew)
macOS 13.5
Swfit: 5.8.1
Xcode: 14.3.1

sidepelican:
Solved with #267

PR #268, opened 2024-10-26
Add test for read-only computed property in class

sidepelican:
The issue in https://github.com/uber/mockolo/issues/245 was secondarily solved by https://github.com/uber/mockolo/pull/267 .

This PR adds tests for the case described in issue #245.


fummicc1 (approved):
LGTM!

PR #269, opened 2024-10-28
Fix to avoid using global actors as default values

### LINKED ISSUE #171: Actor annotated protocols don't work
If you annotate a protocol with a global actor, such as @MainActor, its generate mock doesn't compile:

Given
```
/// @mockable
@MainActor
protocol RootController: AnyObject {
    var viewController: UIViewController { get }
}
```

the mock will fail with this compilation error:

```


❌ /usr/local/var/buildkite-agent/builds/7003-local-1/varun-santhanam/{}/Project/{}Tests/Mocks.swift:226:16: call to main actor-isolated initializer 'init()' in a synchronous nonisolated context
--
  | return RootControllerMock()
```

The solution, I believe, would be to also add the actor annotation to the Mock

An identically generated mock, with the `@MainActor` annotation would work just fine.

I believe there would be similar errors for per-member annotations as well.

sidepelican:
Fixes https://github.com/uber/mockolo/issues/171 .

This PR includes two changes:

1. Fix for cases where `any` annotated existential types cannot find a default mock value.

Base

```swift
/// @mockable
protocol SomeProtocol { }

/// @mockable
protocol UseSomeProtocol {
    func foo() -> any SomeProtocol
}
```

Before

```swift
class UseSomeProtocolMock: UseSomeProtocol {
    ...
    func foo()  -> any SomeProtocol {
        fooCallCount += 1
        if let fooHandler = fooHandler {
            return fooHandler()
        }
        fatalError("fooHandler returns can't have a default value thus its handler must be set")
    }
}
```

After

```swift
class UseSomeProtocolMock: UseSomeProtocol {
    ...
    func foo()  -> any SomeProtocol {
        fooCallCount += 1
        if let fooHandler = fooHandler {
            return fooHandler()
        }
        return SomeProtocolMock() 
    }
}
```

2. The mock is may have a global actor doesn't provide a default value

Fix #171. Calling a function with a globalActor from a mock is very complicated, so it is easier to give up the default value.


Sources/MockoloFramework/Operations/Generator.swift
```diff
@@ -107,12 +107,21 @@ public func generate(sourceDirs: [String],
...
+            if value.entityNode.mayHaveGlobalActor {
+                return nil
+            }
+            return key.components(separatedBy: "Mock").first
+        },
+        annotatedProtocolMap.filter { !$0.value.entityNode.mayHaveGlobalActor }.map(\.key)
+    ]
+        .flatMap { $0 }
+        .map { typeName in
+            // nameOverride does not work correctly but it giving up.
```
fummicc1, line 121: [nits]

I am sorry, but could you add a meaningful prefix such as FIXME so that we can easily address its problem in the future. What do you think of it?

Sources/MockoloFramework/Utils/TypeParser.swift
```diff
@@ -685,14 +686,10 @@ public final class SwiftType {
          "UIColor": ".black",
          "UIFont": ".systemFont(ofSize: 12)",
          "UIImage": "UIImage()",
-         "UIView": "UIView(frame: .zero)",
-         "UIViewController": "UIViewController()",
-         "UICollectionView": "UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewLayout())",
-         "UICollectionViewLayout": "UICollectionViewLayout()",
-         "UIScrollView": "UIScrollView()",
```
fummicc1, line 692: [question]

These changes aim to disable defaultValue of globalActor, but some users still use swift 5 mode, I think. so these changes might break backward compatibilities, in my understanding.
What do you think of it?

fummicc1 (commented):
Thank you so much! Implementation is nice. Just to confirm, I added some comments.

Sources/MockoloFramework/Operations/Generator.swift
sidepelican: I am not sure I want this fixed.
If you think anything in this comment, it might be better there are no comments.

Sources/MockoloFramework/Utils/TypeParser.swift
sidepelican: This seems to be a breaking change, but imagine users who would be affected by this.
It shouldn't have been compiling correctly in the first place. That's what #171 is about.

The @MainActor attribute was introduced in Swift 5.1, and mockolo currently supports Swift 5.7 and above. There's no need to consider users with Swift versions below 5.1.

fummicc1: I see. I misunderstood what this issue says.

Sources/MockoloFramework/Operations/Generator.swift
fummicc1: I got it. I just concerned if nameOverride support is hard of not. Thank you for your confirmation.

fummicc1 (approved):
LGTM. Thank you!

karim-alweheshy:
This is a breaking change and should have went on a 3.0.0 version instead

PR #270, opened 2024-10-29
Add Nested protocol support

sidepelican:
Swift 5.10 introduces support for nested protocols.
You can read more about this update in the Swift Evolution proposal: [Nested Protocols](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0404-nested-protocols.md).
This PR adds support for generating mocks for these nested protocols.

### Unrelated Changes

- The `Entity.data` property is no longer used, so it has been removed.


fummicc1 (approved):
LGTM from my side. Thank you!

PR #271, opened 2024-11-06
Remove dead codes

sidepelican:
There is a lot of dead code in mockolo due to historical reasons.
With many unused definitions and arguments, they make hard to understand of the code, so I would like to clean up them.

In this PR, I will make some simple deletions and replace some implementations with simpler forms

### DIFF
--- a/Sources/MockoloFramework/Models/ParamModel.swift
+++ b/Sources/MockoloFramework/Models/ParamModel.swift
@@ -17,37 +17,32 @@
 import Foundation
 
 final class ParamModel: Model {
-    var name: String
-    var offset: Int64
-    var length: Int64
-    var type: SwiftType
-    let label: String
-    let isGeneric: Bool
-    let inInit: Bool
-    let needVarDecl: Bool
-
-    var isStatic: Bool {
-        return false
+    internal init(label: String, name: String, type: SwiftType, isGeneric: Bool, inInit: Bool, needsVarDecl: Bool, offset: Int64, length: Int64) {
+        self.label = label
+        self.name = name
+        self.type = type
+        self.isGeneric = isGeneric
+        self.inInit = inInit
+        self.needsVarDecl = needsVarDecl
+        self.offset = offset
+        self.length = length
     }
     
+    let label: String
+    let name: String
+    let type: SwiftType
+    let isGeneric: Bool
+    let inInit: Bool
+    let needsVarDecl: Bool
+    let offset: Int64
+    let length: Int64
+
     var modelType: ModelType {
         return .parameter
     }
 
     var fullName: String {
         return label + "_" + name
-    }
-    
-    init(label: String = "", name: String, typeName: String, isGeneric: Bool = false, inInit: Bool = false, needVarDecl: Bool, offset: Int64, length: Int64) {
-        self.name = name.trimmingCharacters(in: .whitespaces)
-        self.type = SwiftType(typeName.trimmingCharacters(in: .whitespaces))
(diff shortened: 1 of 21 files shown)


Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -30,13 +39,6 @@ extension SyntaxProtocol {
     }
 }
 
-
-extension AttributeListSyntax {
-    var trimmedDescription: String? {
```
sidepelican, line 35: SwiftSyntax already has `trimmedDescription`

Sources/MockoloFramework/Templates/NominalTemplate.swift
```diff
@@ -270,21 +270,14 @@ extension NominalModel {
...
-                }
+        var aliasMap = [String: Set<String>]()
+        for (_, model) in models {
+            if let alias = model as? TypeAliasModel {
+                aliasMap[alias.name, default: []].insert(alias.type.typeName)
             }
         }
         let aliasDupes = aliasMap.filter {$0.value.count > 1}
-        return aliasDupes.isEmpty ? nil : aliasDupes
+        return aliasDupes.isEmpty ? nil : aliasDupes.mapValues {$0.sorted()}
```
sidepelican, line 280: The results are now sorted, but I believe there are no compatibility issues.

Sources/MockoloFramework/Models/ParamModel.swift
```diff
@@ -17,19 +17,26 @@
 import Foundation
 
 final class ParamModel: Model {
-    var name: String
-    var offset: Int64
-    var length: Int64
-    var type: SwiftType
+    internal init(label: String, name: String, type: SwiftType, isGeneric: Bool, inInit: Bool, needsVarDecl: Bool, offset: Int64, length: Int64) {
```
uhooi, line 20: [Question]
Why is it explicitly marked internal?

Sources/MockoloFramework/Operations/UniqueModelGenerator.swift
```diff
@@ -52,13 +51,15 @@ private func generateUniqueModels(key: String,
...
+    let (nonMethodModels, methodModels) = models.partitioned(by: { $0.modelType == .method })
     let orderedModels = [nonMethodModels, methodModels].flatMap {$0}
-    let x = uniqueEntities(in: orderedModels, exclude: processedLookup, fullnames: processedFullNames)
-    let unmockedUniqueEntities = x.filter {!$0.value.processed}
-    
+    let unmockedUniqueEntities = uniqueEntities(
+        in: orderedModels,
+        exclude: processedLookup,
+        fullnames: processedFullNames
+    ).filter {!$0.value.processed}
```
uhooi, line 61: [nits]
I need a space before and after the brace.

```suggestion
    ).filter { !$0.value.processed }
```

uhooi (approved):

Sources/MockoloFramework/Models/ParamModel.swift
sidepelican: It doesn't have any particular meaning. It gets added when I used 'Generate Memberwise Initializer'. I left it as is.

Sources/MockoloFramework/Operations/UniqueModelGenerator.swift
sidepelican: In mockolo's style, it seems that spaces are not used here.

PR #272, opened 2024-11-06
Organize the arguments of Model.render.

sidepelican:
`Model.render` has a lot of arguments, making it very complex.
There are also places where meaningless values are being passed.

In this PR, I will organize the arguments by separating them into `RenderContext` and `GenerationArguments`, allowing us to distinguish between global settings and the state during the render process.

### DIFF
--- a/Sources/MockoloFramework/Models/ClosureModel.swift
+++ b/Sources/MockoloFramework/Models/ClosureModel.swift
@@ -17,7 +17,6 @@
 import Foundation
 
 final class ClosureModel: Model {
-    var type: SwiftType
     let name: String = "" // closure type cannot have a name
     let offset: Int64 = .max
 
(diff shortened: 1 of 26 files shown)


Sources/MockoloFramework/Templates/NominalTemplate.swift
```diff
@@ -50,22 +44,22 @@ extension NominalModel {
...
+                        enclosingType: type,
+                        annotatedTypeKind: declKindOfMockAnnotatedBaseType
+                    ),
+                    arguments: arguments
+                ) {
+                    return ret
                 }
                 return nil
-        }
-        .sorted { (left: (String, Int64), right: (String, Int64)) -> Bool in
```
sidepelican, line 58: The sorting is done in `generateUniqueModels`, so sorting here is unnecessary.

Tests/TestMacros/MacroTests.swift
```diff
@@ -1,35 +1,42 @@
...
-class MacroTests: MockoloTestCase {
-   func testMacroInFunc() {
+final class MacroTests: MockoloTestCase {
+    func testMacroInFunc() {
         verify(srcContent: macroInFunc,
                dstContent: macroInFuncMock)
     }
 
+    func testMacroInFuncWithOverload() {
+        XCTExpectFailure("Resolving overloading in #if is broken.") {
```
sidepelican, line 10: In the process of refactoring, I discovered a new bug🤣

Sources/MockoloFramework/Models/ClosureModel.swift
```diff
@@ -32,28 +31,37 @@ final class ClosureModel: Model {
         return .closure
     }
 
-    init(genericTypeParams: [ParamModel], paramNames: [String], paramTypes: [SwiftType], isAsync: Bool, throwing: ThrowingKind, returnType: SwiftType, encloser: String) {
```
sidepelican, line 35: `encloser` is unnecessary here because `type` depends on `RenderContext`.

Sources/MockoloFramework/Models/Model.swift
```diff
@@ -27,6 +27,18 @@ public enum ModelType {
     case closure
 }
 
+enum NominalTypeDeclKind: String {
```
sidepelican, line 30: In the most places, `DeclType` has incorrect cases so prepares new enum.

Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -131,14 +125,31 @@ struct AnnotationMetadata {
     var combineTypes: [String: CombineType]?
 }
 
+struct GenerationArguments {
```
fummicc1, line 128: [imho]

```suggestion
struct RenderArguments {
```

I might prefer RenderArguments to align word choice with RenderContext, though it is not must.

Tests/TestMacros/MacroTests.swift
fummicc1: New finding! let's create an issue.

fummicc1 (commented):
I left a few minor review comments. Thank you very much!

Sources/MockoloFramework/Models/ParsedEntity.swift
sidepelican: The `GenerationArguments` means the arguments of `func generate`.
The value is unique (fixed) for applying `generate`, unlike the render process.

fummicc1: I could not distinguish the difference. Thank you for your information.🙇‍♂️

fummicc1 (approved):
Thank you very much!

PR #277, opened 2024-11-28
Bump up version to 2.2.0

sidepelican:
prepare new version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,9 +6,9 @@ on:
     types: [published]
 
 env:
-  SWIFT_VERSION: "6.0"
-  SWIFT_SDK_URL: https://download.swift.org/swift-6.0.1-release/static-sdk/swift-6.0.1-RELEASE/swift-6.0.1-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -6,9 +6,9 @@ on:
     types: [published]
 
 env:
-  SWIFT_VERSION: "6.0"
-  SWIFT_SDK_URL: https://download.swift.org/swift-6.0.1-release/static-sdk/swift-6.0.1-RELEASE/swift-6.0.1-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz
-  SWIFT_SDK_CHECKSUM: d4f46ba40e11e697387468e18987ee622908bc350310d8af54eb5e17c2ff5481
+  SWIFT_VERSION: "6.0.2"
```
sidepelican, line 9: It was needed to align swift version and swift-sdk version

sidepelican:
I see https://github.com/uber/mockolo/issues/261#issuecomment-2503521093, and concurrency support may take some time, so I thought it would be a good idea to do a release once.

fummicc1 (approved):
LGTM from my side!

Issue #263, opened 2024-09-30
Compiler directive else is ignored


kaanbiryol:
I saw the issue #83 was implemented however I had some observations. 

Given a source file `source.swift`;

```swift
#if DEBUG
import X
#elseif RELEASE
import Y
#endif

/// @mockable
protocol SwiftTest {}
```

`mockolo -srcs source.swift -d Mock.swift && cat Mock.swift` outputs;

```swift
#if DEBUG
import X
#endif
#if RELEASE
import Y
#endif
```

which is reflected correctly.

However;

```swift
#if DEBUG
import X
#else
import Y
#endif

/// @mockable
protocol SwiftTest {}
```
outputs;

```swift
import X
import Y
#if DEBUG
import X
#endif
```

which is incorrect. `import Y` should only be there for `#else` case.

Another observation (I don't know if this is supported or not) I had was compiler directives are not working inside the `protocol` definitions.

```swift
/// @mockable
protocol SwiftCompilerTest {
    #if compiler(<4.0)
    func swiftLessThan4()
    #else
    func swift()
    #endif
}
```

outputs;

```swift
class SwiftCompilerTestMock: SwiftCompilerTest {
    init() { }

    #if compiler(<4.0)

    private(set) var swiftLessThan4CallCount = 0
    var swiftLessThan4Handler: (() -> ())?
    func swiftLessThan4()  {
        swiftLessThan4CallCount += 1
        if let swiftLessThan4Handler = swiftLessThan4Handler {
            swiftLessThan4Handler()
        }

    }
    #endif
}
```

`func swift()` seems to be completely ignored in this case.

- mockolo version : `2.1.1` 
- swift version: `swift: swiftlang-6.0.0.9.10 clang-1600.0.26.2`

PR #279, opened 2024-12-01
Avoid duplicated variable under if-macro syntax

### LINKED ISSUE #273: overload inside if macro produces same variables
Overload inside `#if` macro produces same handlers that results compile error.

- Input

```swift
/// @mockable
protocol PresentableListener: AnyObject {
    #if DEBUG
    func run(value: Int)
    func run(value: String)
    #endif
    func run2(value: Int)
    func run2(value: String)
}
```

- Output (actual)

```swift

class PresentableListenerMock: PresentableListener {
    init() { }

    #if DEBUG

    private(set) var runCallCount = 0
    var runHandler: ((Int) -> ())?
    func run(value: Int) {
        runCallCount += 1
        if let runHandler = runHandler {
            runHandler(value)
        }
        
    }

    // not supporting overload
    private(set) var runCallCount = 0
    var runHandler: ((String) -> ())?
    func run(value: String) {
        runCallCount += 1
        if let runHandler = runHandler {
            runHandler(value)
        }
        
    }
    #endif

    private(set) var run2CallCount = 0
    var run2Handler: ((Int) -> ())?
    func run2(value: Int) {
        run2CallCount += 1
        if let run2Handler = run2Handler {
            run2Handler(value)
        }
        
    }

    // supporting overload
    private(set) var run2ValueCallCount = 0
    var run2ValueHandler: ((String) -> ())?
    func run2(value: String) {
        run2ValueCallCount += 1
        if let run2ValueHandler = run2ValueHandler {
            run2ValueHandler(value)
        }
        
    }
}


```

- Expected output

mock vars for second func `run(value:)` inside if macro should be `runValueCount` and `runValueHandler`.

fummicc1:
This PR addresses #273 and performs `uniqueEntities` method for IfMacroModel before rendering because entities inside IfMacroModel are not treated as NominalModel's entities.

I think IfMacro support is insufficient yet (see #263  ) but I am focusing to fix #273 in this PR.

- Input

```swift
/// @mockable
protocol PresentableListener: AnyObject {
    #if RELEASE
    func run(value: Int)
    func run(value: String)
    #endif
}
```

- Output

<table>
<thead><tr><th>Before</th><th>After</th></tr></thead>
<tbody>
<tr><td>

```swift
class PresentableListenerMock: PresentableListener {
    init() { }

    #if RELEASE

    private(set) var runCallCount = 0
    var runHandler: ((Int) -> ())?
    func run(value: Int) {
        runCallCount += 1
        if let runHandler = runHandler {
            runHandler(value)
        }
        
    }

    // not supporting overload
    private(set) var runCallCount = 0
    var runHandler: ((String) -> ())?
    func run(value: String) {
        runCallCount += 1
        if let runHandler = runHandler {
            runHandler(value)
        }
        
    }
    #endif
}
```

</td><td>

```swift

class PresentableListenerMock: PresentableListener {
    init() { }

    #if RELEASE

    private(set) var runCallCount = 0
    var runArgValues = [Int]()
    var runHandler: ((Int) -> ())?
    func run(value: Int) {
        runCallCount += 1
        runArgValues.append(value)
        if let runHandler = runHandler {
            runHandler(value)
        }
        
    }

    private(set) var runValueCallCount = 0
    var runValueArgValues = [String]()
    var runValueHandler: ((String) -> ())?
    func run(value: String) {
        runValueCallCount += 1
        runValueArgValues.append(value)
        if let runValueHandler = runValueHandler {
            runValueHandler(value)
        }
        
    }
    #endif
}
``` 

</td></tr>
</tbody></table>

### DIFF
--- a/Sources/MockoloFramework/Models/IfMacroModel.swift
+++ b/Sources/MockoloFramework/Models/IfMacroModel.swift
@@ -17,19 +17,19 @@
 final class IfMacroModel: Model {
     let name: String
     let offset: Int64
-    let entities: [Model]
+    let entities: [(String, Model)]
 
     var modelType: ModelType {
         return .macro
     }
 
     var fullName: String {
-        return entities.map {$0.fullName}.joined(separator: "_")
+        return entities.map {$0.0}.joined(separator: "_")
     }
     
     init(name: String,
          offset: Int64,
-         entities: [Model]) {
+         entities: [(String, Model)]) {
         self.name = name
         self.entities = entities
         self.offset = offset
--- a/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
+++ b/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
@@ -254,8 +254,14 @@ extension IfConfigDeclSyntax {
                 }
             }
         }
+        
+        let uniqueSubModels = uniqueEntities(
+            in: subModels,
+            exclude: [:],
+            fullnames: []
+        ).map({ $0 })
 
-        let macroModel = IfMacroModel(name: name, offset: self.offset, entities: subModels)
+        let macroModel = IfMacroModel(name: name, offset: self.offset, entities: uniqueSubModels)
         return (macroModel, attrDesc, hasInit)
     }
 }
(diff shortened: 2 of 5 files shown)


sidepelican (approved):
This change is good, but it's not perfect.
The same issue will continue to occur with protocols like the following:

```swift
/// @mockable
protocol PresentableListener {
    #if DEBUG
    func run(value: Int)
    #endif
    func run(value: Double)
}
```

But because it makes things better than they are now, I think it is best to merge this change for now.

PR #280, opened 2024-12-03
Make fixture verification a bit strict

sidepelican:
Currently, in tests, the comparison between the fixture and the actually generated code is done on a word-by-word basis, separated by whitespace.
There might be unnecessary whitespace present in the actually generated code, which is not being detected.

I will make the process slightly more strict and change it to compare on a line-by-line basis.

### DIFF
--- a/Tests/TestEmojis/FixtureEmojis.swift
+++ b/Tests/TestEmojis/FixtureEmojis.swift
@@ -51,7 +51,7 @@ class EmojiVarsMock: EmojiVars {
 
 
     private(set) var 😂SetCallCount = 0
-    private var _😂: Emoji!  { didSet { 😂SetCallCount += 1 } }
+    private var _😂: Emoji! { didSet { 😂SetCallCount += 1 } }
     var 😂: Emoji {
         get { return _😂 }
         set { _😂 = newValue }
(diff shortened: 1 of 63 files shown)


Tests/TestInit/FixtureInit.swift
```diff
@@ -307,11 +298,8 @@ protocol MyProtocol {
 """
 
 let multipleInitsWithSameParamNameMock = """
-
-
-
 class MyProtocolMock: MyProtocol {
-        private var _anotherParam: Any!
+    private var _anotherParam: Any!
```
fummicc1, line 302: Improvement of readability🚀

Tests/TestEmojis/FixtureEmojis.swift
```diff
@@ -161,16 +158,14 @@ class FamilyMock: Family {
 """
 
 
-let krJp =
-"""
+let krJp = """
 /// \(String.mockAnnotation)
 protocol Hello: Hi {
-    var 天気が: String { get set }
```
fummicc1, line 168: [question]

Is there any reason that removes `が`?

fummicc1 (approved):
Thank you so much! LGTM from my side. 
This fix improves readability of generated mocks and makes it easy to maintain test code

PR #281, opened 2024-12-05
Fix undetectable mock annotations with attributes above.

sidepelican:
mockolo cannot detect the mock annotation in code like the following:

```swift
@MainActor
/// @mockable
protocol P {
}
```
This PR fixes this issue.

### DIFF
--- a/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
+++ b/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
@@ -259,7 +259,7 @@ extension IfConfigDeclSyntax {
             in: subModels,
             exclude: [:],
             fullnames: []
-        ).map({ $0 })
+        ).sorted(path: \.value.offset, fallback: \.key)
 
         let macroModel = IfMacroModel(name: name, offset: self.offset, entities: uniqueSubModels)
         return (macroModel, attrDesc, hasInit)
--- a/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
+++ b/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
@@ -300,7 +300,7 @@ extension ProtocolDeclSyntax: EntityNode {
     }
 
     func annotationMetadata(with annotation: String) -> AnnotationMetadata? {
-        return leadingTrivia.annotationMetadata(with: annotation)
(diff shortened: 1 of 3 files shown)


Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -259,7 +259,7 @@ extension IfConfigDeclSyntax {
             in: subModels,
             exclude: [:],
             fullnames: []
-        ).map({ $0 })
+        ).sorted(path: \.value.offset, fallback: \.key)
```
sidepelican, line 262: Additionally, fix the fragile test caused by https://github.com/uber/mockolo/pull/279 .

sidepelican: This operation is same as  https://github.com/uber/mockolo/blob/1fb44ed25e962b52e18f293201a192375f9e018a/Sources/MockoloFramework/Operations/UniqueModelGenerator.swift#L69 .

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -300,7 +300,7 @@ extension ProtocolDeclSyntax: EntityNode {
     }
 
     func annotationMetadata(with annotation: String) -> AnnotationMetadata? {
-        return leadingTrivia.annotationMetadata(with: annotation)
+        return leadingTrivia.annotationMetadata(with: annotation) ?? protocolKeyword.leadingTrivia.annotationMetadata(with: annotation)
```
fummicc1, line 303:
```suggestion
        let trivias = [
            leadingTrivia,
            protocolKeyword.leadingTrivia,
            modifiers.leadingTrivia
        ] + attributes.map(\.leadingTrivia)
        return trivias.lazy.compactMap { $0.annotationMetadata(with: annotation) }.first
```

I found this is not sufficient. Below is a failure pattern.

```swift
@MainActor
/// @mockable
public protocol P {
    func f() -> String
}
```

fummicc1 (changes requested):
Thank you very much! I found one case that should be supported.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
sidepelican: Good catch! thank you.

fummicc1 (approved):
LGTM! Thank you for fixing fragile test as well. :)

PR #282, opened 2024-12-09
Compile fixture code with macro

sidepelican:
Currently, there are many Fixtures in the mockolo test code, but they are not statically checked by the compiler.
As a result, incorrect code may be tested.

In this PR, we introduce the `@Fixture` macro and change the tests to use actual Swift code.
The real Swift code will be validated for correctness at compile time, and the input and output of the tests will use the source code strings generated by the macro.


sidepelican:
We don't replace all the tests in this PR. To validate the approach, only some of the tests have been updated.

sidepelican:
The following two test cases have been commented out because they were incorrect (they won't be fixed in this PR):
- variadicFunc
- closureArgFunc

fummicc1 (approved):
Sorry for the late review but LGTM from my side!

Issue #288, opened 2025-02-25
Unable to correctly output mock args history of Sendable protocol


slightair:
## Environment

- Xcode 16.2
- mockolo v2.3.0

## Issue

When defining a method with multiple arguments in a protocol conforming to Sendable, build errors occur in the generated code for args history.

The following input is slightly added to the contents of FixtureSendable.swift.
The generated code for `func update(arg: Int) -> String` works as expected.

input:
```swift
/// @mockable
public protocol SendableProtocol: Sendable {
    func update(arg: Int) -> String
    func update(arg0: Int, arg1: Int) -> String
    func update(arg0: some Sendable, arg1: AnyObject) async throws
}
```

output:
```swift
...
    public func update(arg0: Int, arg1: Int) -> String {
        warnIfNotSendable(arg0, arg1)
        let updateArg0Handler = updateArg0State.withLock { state in
            state.callCount += 1
            state.argValues.append(.init(arg0, arg1)) // <-- error: initializer 'init(_:)' requires the types '(arg0: Int, arg1: Int)' and '(Int, Int)' be equivalent
            return state.handler
        }
        if let updateArg0Handler = updateArg0Handler {
            return updateArg0Handler(arg0, arg1)
        }
        return ""
    }
...
    public func update(arg0: some Sendable, arg1: AnyObject) async throws {
        warnIfNotSendable(arg0, arg1)
        let updateArg0Arg1Handler = updateArg0Arg1State.withLock { state in
            state.callCount += 1
            state.argValues.append(.init(arg0, arg1)) // <-- error: initializer 'init(_:)' requires the types '(arg0: Any, arg1: AnyObject)' and '(some Sendable, AnyObject)' be equivalent
            return state.handler
        }
        if let updateArg0Arg1Handler = updateArg0Arg1Handler {
            try await updateArg0Arg1Handler(arg0, arg1)
        }
        
    }
...
```

sidepelican:
Thanks for reporting issue!
This was fixed in 2.3.1

PR #289, opened 2025-02-26
Fix compile error in Sendable mock with labeled history

sidepelican:
- https://github.com/uber/mockolo/issues/288

Normally, tuples are considered to be of the same type as long as their shape is the same, regardless of whether they have labels.
However, parameter packed tuples are different; they are always regarded as mismatched types with labeled tuples.
I'm not sure if this is intentional behavior by the Swift compiler, but as it stands, it results in a compilation error and caused the issue in #288.

To fix this, I stop using parameter packs.

Additionally, this issue could have been detected earlier if `FixtureSendable.swift` had been compiled.
I have rewritten the fixture using the `@Fixture` macro to ensure it compiles.

I want to release this change as 2.3.1.


Sources/MockoloFramework/Templates/NominalTemplate.swift
```diff
@@ -257,9 +257,14 @@ extension NominalModel {
         }
 
         if needBlankInit {
-            // In case of protocol mocking, we want to provide a blank init (if not present already) for convenience,
-            // where instance vars do not have to be set in init since they all have get/set (see VariableTemplate).
-            let blankInit = "\(acl)init() { }"
+            let blankInit: String
+            if context.annotatedTypeKind == .class {
+                blankInit = "\(acl)override init() { }"
```
sidepelican, line 262: Compiling `FixtureSendable.swift` I found another issue, which has been fixed as well.

Tests/TestSendable/FixtureSendable.swift
```diff
@@ -1,190 +1,135 @@
...
-        get { updateState.withLock(\.handler) }
-        set { updateState.withLock { $0.handler = newValue } }
-    }
-    public func update(arg: Int) -> String {
-        let updateHandler = updateState.withLock { state in
-            state.callCount += 1
-            return state.handler
+    @Fixture enum expected {
+        public final class UncheckedSendableClassMock: UncheckedSendableClass, @unchecked Sendable {
+            public override init() { }
```
sidepelican, line 76: I found the lack of `override` here.

Tests/TestConcurrencyHelpers.swift
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func warnIfNotSendable<each T>(function: String = #function, _: repeat each T) {
+    print("At \(function), the captured arguments are not Sendable, it is not concurrency-safe.")
+}
+func warnIfNotSendable<each T: Sendable>(function: String = #function, _: repeat each T) {
+}
+typealias MockoloMutex = TestConcurrencyHelpers.MockoloMutex
+typealias MockoloUnsafeTransfer = TestConcurrencyHelpers.MockoloUnsafeTransfer
+typealias MockoloHandlerState = TestConcurrencyHelpers.MockoloHandlerState
```
sidepelican, line 54: It became necessary to use these functions within the test fixtures, so I am ensuring here that they match what mocokolo generates and making them available for use within the tests.

fummicc1: I'm sorry to miss the compile error and thank you for adding a new test case. 
(Fixture macro is really helpful to check if test-fixture is validity or not.)

Tests/TestConcurrencyHelpers.swift
```diff
@@ -0,0 +1,83 @@
...
+
+import XCTest
+
+final class TestConcurrencyHelpersTests: MockoloTestCase {
```
fummicc1, line 58:
```suggestion
final class ConcurrencyHelpersTests: MockoloTestCase {
```

no hurry and nits.
In my opinion, a system in this test case is not TestConcurrencyHelper but [ConcurrencyHelper](https://github.com/sidepelican/mockolo/blob/2258ee1b13347eaf56755451703fcb0b9bc8f085/Sources/MockoloFramework/Templates/ConcurrencyHelpersTemplate.swift#L1). so this class might be `ConcurrencyHelpersTests` if I understand correctly.

fummicc1 (approved):
LGTM! I added nits comment.

PR #291, opened 2025-03-13
Use @Fixture in VarTests

sidepelican:
this PR simply rewrite `VarTests` to use `@Fixture` macro.


Tests/TestVars/FixtureNonSimpleVars.swift
```diff
@@ -1,50 +1,49 @@
 import MockoloFramework
 
-let nonSimpleVars = """
-import Foundation
-
-/// @mockable
-@objc
```
sidepelican, line 7: This `@objc` annotation has been removed for linux build. That didn't have meanings in this test.

PR #292, opened 2025-03-16
Remove `metadata` from NominalModel, resulting correct Rx SubjectType in mock's init.

sidepelican:
`NominalModel` has `metadata`, but I've realized that this annotation should not be used after the model is created. I'm making changes so that `NominalModel` no longer needs to use `metadata` later on.
When a protocol inherits from another protocol, the `metadata` implicitly acts as if it is inherited. However, if `metadata` is processed after the model is created, the inheritance is not handled correctly.

### DIFF
--- a/Sources/MockoloFramework/Models/Model.swift
+++ b/Sources/MockoloFramework/Models/Model.swift
@@ -38,6 +38,7 @@ struct RenderContext {
     var enclosingType: SwiftType?
     var annotatedTypeKind: NominalTypeDeclKind?
     var requiresSendable: Bool = false
+    var metadata: AnnotationMetadata?
 }
 
 /// Represents a model for an entity such as var, func, class, etc.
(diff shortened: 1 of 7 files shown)


fummicc1 (approved):
LGTM from my side!

This PR also corrects Rx Subject type in init default value for mock of nominal model (though only when its parent overrides `rx` in mockolo annotation), if I understand correctly.
Do you think this change needs to be included in PR title or description like the following so that we can pick it up in next release note?

```
[NominalModel] Remove `metadata` and correct Rx Subject type in init.
```

sidepelican:
> You are trying to fix it and will make succeeding PR?

The problem was already fixed in this PR. The problem was the incorrect Rx SubjectType.

fummicc1:
> The problem was already fixed in this PR. The problem was the incorrect Rx SubjectType.

I got it!

PR #294, opened 2025-03-25
When associatedtype has constraints, mock class becomes generic

sidepelican:
## Background

While I was rewriting `FixturePAT.swift` using `@Fixture`, I noticed that the implementation related to associatedtype was incorrect.

```swift
/// @mockable
protocol Foo {
    associatedtype U: Collection where U.Element == Int
}

class FooMock: Foo {
    // 🆖 'where' clause on non-generic member declaration requires a generic context
    typealias U = Collection where U.Element == Int
}
```

```swift
protocol Foo {
    associatedtype T = Int
}

protocol Bar {
    associatedtype T: StringProtocol
}

/// @mockable
protocol Baz: Foo, Bar {
}

class BazMock: Baz {
    // 🆖 Int is not StringProtocol
    typealias T = Int
}
```

Since mockolo cannot evaluate constraints attached to generic types, I believe that for protocols with generic type constraints, the mock type should also have a generic type parameter.

## Solution

Enable mock types to have generic type parameters.

Introduce `AssociatedTypeModel` and consider it during rendering to determine whether to include a generic type parameter or prepare a typealias.
Add `genericWhereConstraints` to `NominalModel` and make a similar judgment accordingly.

### DIFF
--- a/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
+++ b/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
@@ -183,7 +183,7 @@ extension MemberBlockItemSyntax {
             }
         } else if let patMember = self.decl.as(AssociatedTypeDeclSyntax.self) {
             let acl = memberAcl(patMember.modifiers, encloserAcl, declKind)
-            let item = patMember.model(with: acl, declKind: declKind, overrides: metadata?.typeAliases, processed: processed)
+            let item = patMember.model(with: acl, declKind: declKind, overrides: metadata?.typeAliases)
             return (item, patMember.attributes.trimmedDescription, false)
         } else if let taMember = self.decl.as(TypeAliasDeclSyntax.self) {
             let acl = memberAcl(taMember.modifiers, encloserAcl, declKind)
@@ -645,19 +645,17 @@ extension FunctionParameterSyntax {
 }
 
 extension AssociatedTypeDeclSyntax {
-    func model(with acl: String, declKind: NominalTypeDeclKind, overrides: [String: String]?, processed: Bool) -> Model {
+    func model(with acl: String, declKind: NominalTypeDeclKind, overrides: [String: String]?) -> Model {
         // Get the inhertied type for an associated type if any
         var t = self.inheritanceClause?.typesDescription ?? ""
         t.append(self.genericWhereClause?.description ?? "")
 
-        return TypeAliasModel(name: self.name.text,
-                              typeName: t,
-                              acl: acl,
-                              overrideTypes: overrides,
-                              offset: self.offset,
-                              length: self.length,
-                              modelDescription: self.description,
-                              processed: processed)
+        return AssociatedTypeModel(name: self.name.text,
+                                   typeName: t,
+                                   acl: acl,
+                                   overrideTypes: overrides,
+                                   offset: self.offset,
+                                   length: self.length)
     }
 }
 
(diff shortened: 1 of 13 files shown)


Tests/TestPATs/FixturePAT.swift
```diff
@@ -1,162 +1,235 @@
 import MockoloFramework
 
-let patNameCollision = """
-/// @mockable
-protocol Foo {
-associatedtype T
-}
+@Fixture enum simplePat {
+    protocol Foo {}
```
fummicc1, line 4:
```suggestion
    /// @mockable
    protocol Foo {}
```

fummicc1 (commented):
Thank you! I am still reviewing 🙏

Tests/TestPATs/FixturePAT.swift
```diff
@@ -1,162 +1,235 @@
...
+            // typealias U = Int
+        }
+
+        class QuxMock<T>: Qux where T: Collection {
+            init() { }
+
+            // Unavailable due to the presence of generic constraints
+            // typealias T = Int
+        }
+    }
```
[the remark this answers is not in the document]
sidepelican: There is a bug regarding the use of associatedtype and type parameters in swift 5.10.

```swift
enum E {
  protocol P where T: Collection {
    associatedtype T = Int
  }
  struct S<T>: P where T: Collection {
  }
}
```

This code can be compiled on swift 6 but not on swift 5.10.

Tests/TestPATs/FixturePAT.swift
sidepelican: This is ok because `FooMock` is provided in "parent mock".
(I'm not sure what the original test want to ensure)

sidepelican: "parent mock" is a crazy complicated feature of mockolo. I cannot describe the complete feature about it and it is not documented.

fummicc1: Thank you for your explanation. I misunderstood.

Tests/TestPATs/FixturePAT.swift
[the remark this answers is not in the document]
fummicc1: I got it. Thank you!

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -296,6 +292,10 @@ extension ProtocolDeclSyntax: EntityNode {
         return inheritanceClause?.types ?? []
     }
 
+    var genericWhereConstraints: [String] {
```
fummicc1, line 295: [IMHO]

When I initially saw this variable, I couldn't clearly understand what value would return.

To improve readability, how about making sort of GenericWhereConstraints class like the following?

```swift

// where P: C
struct GenericsWhereConstraint {
    let left: String // P
    let right: String // C
}
```

fummicc1 (approved):
LGTM with nits comment! I am sorry for late reviewing :pray:

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
sidepelican: It would be a re-implementing `GenericRequirementSyntax`, so I think it goes against the role of the `Models` to convert SwiftSyntax nodes into a struct that is easy to use for code generation.

Issue #293, opened 2025-03-24
[chore] Add external dependency library to TestTarget


fummicc1:
## Motivation

Since introduction of `Fixture` macro, we can write robust test fixture. However some of test case cannot apply it due to the dependency of external library such as RxSwift.

## What / How

To resolve the motivation and we can write any test fixture with `Fixture` macro, I suggest adding external library to `MockoloTests` like the following.

```swift

let package = Package(
    name: "Foo",
    products: [
        .library(
            name: "Foo",
            targets: ["Foo"]),
    ],
    dependencies: [
        .package(url: "https://github.com/ReactiveX/RxSwift", .upToNextMajor(from: "6.9.0"))
    ],
    targets: [
        .target(
            name: "Foo",
            dependencies: [
                // unless we write below code, user of mockolo does not resolve RxSwift dependency, in my understanding.
                // .product(name: "RxSwift", package: "RxSwift"), 
            ]
        ),
        .testTarget(
            name: "FooTests",
            dependencies: [
                "Foo",
                .product(name: "RxSwift", package: "RxSwift")
            ]
        ),
    ]
)

```

## Pros

- Write robust and readable test about RxSwift

## Cons

- Need to discuss which RxSwift version to use and how maintain it.
  - As for Combine, I believe we don't have to consider which version are used.

Currently, external dependencies are only RxSwift and Combine. I think we can also continue using legacy test fixture for them.

sidepelican:
I think adding RxSwift as a dependency is a good way to improve the quality of tests.
However I don’t think we should do it because the problems are bigger than the gains.

- It increases compile time and also extends the time Xcode takes to prepare packages.
- RxSwift is mature enough. I think it’s enough to write RxSwift types in `Types.swift`.

https://github.com/uber/mockolo/blob/aa96917c854381d6074cf8cacda39c88b6ce4c54/Tests/Types.swift#L1-L3

- I don’t have a positive opinion about mockolo’s RxSwift-related features. I don’t want to invest effort here.

fummicc1:
I am going to close this issue because RxTests is written with Fixture macro and other tests depending on third party api could be replaced by same manner.

PR #297, opened 2025-04-12
Use `@Fixture` in Rx Tests and remove `use-mock-observable` flag

fummicc1:
## Why

`@Fixture` macro facilitates writing robust tests. However, sut which depends on third party library like `RxSwift` makes it hard to use Fixture macro because of compilation error. ( #293  )

## What

- Created dummy types of Rx.
- Use `@Fixture` macros for Tests.
- Delete `--use-mock-observable` flag from mockolo command line because `MockObservable` property wrapper has not worked since initial implementation.


Sources/MockoloFramework/Templates/PropertyMockable.swift
fummicc1, line 1: I couldn't leave diff well. please check the below.

diff: [`263cd04`](https://github.com/uber/mockolo/pull/297/commits/263cd049b2709d845e98591873bc35caf9aa6279)

Sources/MockoloSupportRxSwift/PropertyMockable.swift
fummicc1, line 1: I couldn't leave diff well. please check the below.

diff: [263cd04](https://github.com/uber/mockolo/pull/297/commits/263cd049b2709d845e98591873bc35caf9aa6279)

Sources/MockoloFramework/Templates/PropertyMockable.swift
```diff
@@ -1,121 +0,0 @@
...
-        } else if let val = newValue as? BehaviorSubject<Value.E> {
-            behaviorSubject = val
-            whichKind = 2
-        } else {
-            fallback = newValue
-            whichKind = 3
-        }
-    }
-}
-
```
fummicc1, line 119: I deleted MockObservableInternal because it is not used anywhere.

fummicc1:
If I understand correctly, RxSwift.ObservableType does not have associatedType `E`, instead it has one named `Element`.

https://github.com/uber/mockolo/blob/2fd3545bd025adbb1ad3e208869d2f59df149385/Sources/MockoloFramework/Templates/PropertyMockable.swift#L37-L42

So I guess there is no user using `MockObservable` feature.

Sources/MockoloSupportRxSwift/PropertyMockable.swift
```diff
@@ -0,0 +1,101 @@
...
+/**
+ The following property wrappers may be used for RxSwift observable variables to be mocked.
+
+ E.g.
+ public class FooMock: Foo {
+    @MockObservable  public var bazStream: Observable<Int> { get }
+ }
+ */
+
+#if canImport(RxSwift)
```
sidepelican, line 26: `canImport` is not work for user defined packages. This is for platform dependent packages.

Sources/MockoloFramework/Templates/PropertyMockable.swift
```diff
@@ -1,121 +0,0 @@
-//
```
sidepelican, line 1: This file is initially broken. `MockoloFramework` is not expected to be used from the test runtime.

There are no ways to use the implementations in this file.

sidepelican (changes requested):
Thank you for the PR. Mockolo is moving forward with stricter tests.
I have a few requests:

- `MockObservable` is initially broken. Please leave it as is (do not touch it) or clean it up completely.
- After that, `MockoloSupportRxSwift` will no longer be meaningful. Remove the target and move any remaining implementations to the test target.

fummicc1:
Thank you for reviewing.

> `MockObservable` is initially broken. Please leave it as is (do not touch it) or clean it up completely.

I can't come up with good idea to make Rx's test-fixture robust without removing MockObservable completely. So I will remove it completely.
Then, the concern is if the change requires to update major version due to breaking change.

> After that, `MockoloSupportRxSwift` will no longer be meaningful. Remove the target and move any remaining implementations to the test target.

I am going to remove `MockoloSupportRxSwift` as well and write dummy Rx types inside test target.

Tests/TestRx/FixtureRxVars.swift
```diff
@@ -1,708 +1,557 @@
...
-    var _integerStream: Observable<Int>? { didSet { integerStreamSubjectSetCallCount += 1 } }
-    var integerStreamSubject = ReplaySubject<Int>.create(bufferSize: 1) { didSet { integerStreamSubjectSetCallCount += 1 } }
-    var integerStream: Observable<Int> {
-        get { return _integerStream ?? integerStreamSubject }
-        set { if let val = newValue as? ReplaySubject<Int> { integerStreamSubject = val } else { _integerStream = newValue } }
-    }
-}
 
+    // Need to use the same name as the parent class to avoid name collision
+    typealias BarMock = parent.BarMock
```
fummicc1, line 24: This typealias is workaround after moving to strict testing. I think this is okay.

sidepelican (approved):
Thank you for your work.

> I will remove it completely.

Nice decision !

> Then, my concern is if the change requires to update major version due to breaking change.

This is certainly destructive change, but it is ok as a minor change, because even though this feature has been broken for a long time, no one was troubled by this.

PR #299, opened 2025-04-15
Avoid Special Handling of the "Unknown" Type Name

sidepelican:
Fix the handling of the type name "Unknown".
Previously, using "Unknown" as a type name would trigger special treatment and lead to unintended behavior.

Stop using empty value as `SwiftType`, it has implicit behavior depending its value. Use void type when representing empty return value.

### DIFF
--- a/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
+++ b/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
@@ -531,7 +531,7 @@ extension FunctionDeclSyntax {
         let genericWhereClause = self.genericWhereClause?.description
 
         let funcmodel = MethodModel(name: self.name.description,
-                                    typeName: self.signature.returnClause?.type.description ?? .voidType,
+                                    typeName: self.signature.returnClause?.type.description,
                                     kind: .funcKind,
                                     acl: acl,
                                     genericTypeParams: genericTypeParams,
@@ -575,7 +575,7 @@ extension InitializerDeclSyntax {
         let genericWhereClause = self.genericWhereClause?.description
 
         return MethodModel(name: "init",
-                           typeName: .voidType,
+                           typeName: nil,
                            kind: .initKind(required: requiredInit, override: declKind == .class),
                            acl: acl,
                            genericTypeParams: genericTypeParams,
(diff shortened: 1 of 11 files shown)


Sources/MockoloFramework/Templates/ParamTemplate.swift
```diff
@@ -20,23 +20,28 @@ extension ParamModel {
...
-                            inInit: Bool) -> String {
+                            inInit: Bool,
+                            isGeneric: Bool) -> String {
         var result = name
         if !label.isEmpty {
             result = "\(label) \(name)"
         }
-        if !type.isUnknown {
+        if isGeneric {
+            // FIXME: `type` is used as a generic constraint
```
sidepelican, line 30: This is caused by wrong `ParamModel` architecture, but fixing this is outside the scope of this PR.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -430,8 +430,8 @@ extension VariableDeclSyntax {
 
         // Need to access pattern bindings to get name, type, and other info of a var decl
         let varmodels = self.bindings.compactMap { (v: PatternBindingSyntax) -> Model in
-            let name = v.pattern.firstToken(viewMode: .sourceAccurate)?.text ?? String.unknownVal
-            var typeName = ""
+            let name = v.pattern.trimmedDescription
+            var typeName = String.voidType
```
fummicc1, line 434: [IMHO]

I would like to use other name for this `typeName` because `void` is sort of reserved word in Swift and this `typeName` is not same as `void` type in Swift.

I would like to suggest alternative naming such as `String.invalidType` or `String.unknownType` for VarDecl typeName.
How do you think of it?

fummicc1 (commented):
LGTM about method side! Let me ask a question about VarDecl.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
sidepelican: You are correct.
It's uncommon but not incorrect for the type name to not appear in variable decl.
Previously, these cases were implicitly treated as a `""` type but I've decided to eliminate this with this PR, so it should explicitly be represented as Optional.

I found similar instances in the `init()` as well, so I've made the same corrections there.

fummicc1: I think representing them as Optional is better approach than previous one. Thank you!

fummicc1 (approved):
LGTM!

PR #300, opened 2025-04-16
Remove `cast` property from `SwiftType`

sidepelican:
The `SwiftType` has a `cast` property, which is a temporary value used only in certain situations and need not be stored.
Remove it because it is structurally unnatural.


fummicc1 (approved):
LGTM!

fummicc1 (approved):
LGTM!

Issue #295, opened 2025-04-01
[Test] unintentional successful tests might happen.


fummicc1:
## What

In MockoloTests, some incorrect outputs may be unintentionally accepted when the expected fixture forgets to include a mock type.

## Example

- Input

```swift
/// @mockable
public protocol Foo {}

/// @mockable(typealias: T = String)
public protocol FooBar<T>: Foo {
  associatedtype T: StringProtocol
}
```

When the above fixture is provided as input, the following output is expected:

```swift
public class FooMock: Foo {
    public init() { }
}

public class FooBarMock: FooBar {
    public init() { }

    public typealias T = String
}
```

However, if the expected fixture forgets to include FooMock, the test case still passes unintentionally:

```swift
// This fixture also passes.
@Fixture enum expected {
   public class FooBarMock: FooBar {
     public init() { }

     public typealias T = String
   }
}
```

This issue occurs because the current assertion only checks whether the expected fixture contains the generated mock code. ([see here](https://github.com/uber/mockolo/blob/aa96917c854381d6074cf8cacda39c88b6ce4c54/Tests/MockoloTestCase.swift#L142))

## How

TBD

sidepelican:
This structure was introduced in https://github.com/uber/mockolo/pull/276/files#r1855429669.

The trigger was the addition of helper implementations for Concurrency.
As background, `GeneratedMocks.swift` typically includes header comments and import statements.
Before #276, these "extra elements that need to be written every time during test inspection" were automatically added to the Fixture itself.
Since these elements were always added, there was no issue with the assert function automatically including them.
However, ConcurrencyHelper was different. Because it dynamically changes depending on whether it’s needed by the mock class, the approach of having the assert function automatically add it became difficult.

To resolve this, we adopted the current structure where "it’s fine as long as a subset of the Fixture is included in `GeneratedMocks.swift`," loosening the verification.
As a side effect, it became harder to detect mistakes in the Fixture file where a mock class that should have been output wasn’t.
That’s what’s being explained in this Issue.

As countermeasures, I can think of two options:
 
1. Accept this specification as it is.  

2. Add whether ConcurrencyHelper should be included as an option in the assert function, and modify the structure so that the assert function automatically appends ConcurrencyHelper.

What do you think?

fummicc1:
Sorry for my late reply.
I believe current method is not best because of insufficient verification, so I might prefer second approach. Thank you for your suggestion. 
After reviewing your Pull Request, I think it's awesome.

PR #304, opened 2025-04-26
Better test verification

sidepelican:
## Issue

- https://github.com/uber/mockolo/issues/295
    - Code that is not included in the Fixture is not being validated.
    - When checking for an exact match with the Fixture, it is necessary to include ConcurrencyHelper in the Fixture each time.
- Other test improvements.

## Solution
- Modify the `@Fixture` macro to make it easy to include ConcurrencyHelper.

## Changes
- Add an option to the `@Fixture` macro so that ConcurrencyHelper can be included.
- Also allow import statements to be specified.
- Display diffs for parts that differ from the Fixture.
- Visualize failed test cases.
- Rewrite `TestableImportStatementsTests` using the `@Fixture` macro.

### DIFF
--- a/Tests/MockoloTestCase.swift
+++ b/Tests/MockoloTestCase.swift
@@ -54,7 +54,7 @@ class MockoloTestCase: XCTestCase {
         }
     }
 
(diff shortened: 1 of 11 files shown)


Tests/MockoloTestCase.swift
```diff
@@ -134,11 +134,47 @@ class MockoloTestCase: XCTestCase {
...
+        """.components(separatedBy: .newlines).filter { !$0.isEmpty && !$0.allSatisfy(\.isWhitespace) }
         if fixtureContents.isEmpty {
-            throw XCTSkip("empty fixture")
+            throw XCTSkip("empty fixture", file: file, line: line)
+        }
+
+        let diff = lightDiff(old: outputContents, new: fixtureContents)
+        if !diff.isEmpty {
+            print("output:\n\(output)")
+            XCTFail("diff:\n" + "\(diff.joined(separator: "\n"))", file: file, line: line)
```
sidepelican, line 150: example:

スクリーンショット 2025-04-26 14 33 33

fummicc1 (approved):
Thank you!

PR #308, opened 2025-05-03
Update CI Configuration

sidepelican:
Switch release build to use Swift 6.1 and update outdated dependencies.

I have decided to stop using the Swift SDK for the following reasons:

- The static Linux SDK significantly increases the size of the output binary.
- Distributing the artifact bundle to every Linux environment is not a priority. Previously (before [#265](https://github.com/uber/mockolo/pull/265)), the artifact was only shipped to Ubuntu, and the main reason for switching to the Swift SDK was to improve build times.
    - Ubuntu will continue to be the primary supported Linux distribution. Support for other distributions can be added upon request.
- We can now utilize ARM Linux machines provided by GitHub, making it straightforward and efficient to build ARM binaries.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -14,11 +14,17 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest]
-        swift: ["5.10", "6.0"]
+        swift: ["5.10", "6.0", "6.1"]
     steps:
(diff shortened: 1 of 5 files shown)


sidepelican:
Working status: https://github.com/sidepelican/mockolo/actions/runs/14810693485

fummicc1 (approved):
LGTM! 4 times more shrink artifactbundle! (from 128MB→30MB)

PR #309, opened 2025-05-04
Support elseif directive

fummicc1:


Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -738,42 +751,77 @@ final class EntityVisitor: SyntaxVisitor {
...
+                                    importLines,
+                                ].joined(separator: "\n")
+                            }
                         }
-                        imports[key]?.append(nested.trimmedDescription)
+                        nestedImportLines.append("\n#endif")
+                        nestedImport = .init(
+                            line: nestedImportLines,
+                            compilerDirectiveKey: compilerDirectiveKey
+                        )
```
fummicc1, line 820: Here, I aggregate all imports inside the compiler directive and wrap them by clause (such as poundIf, poundElseIf and poundElse).

I think this approach is not best, however I believe model architecture refinement might be required to fix underlying issue.

Fortunately, import statement cannot be written in except for top-level. So I think this approach will not produce bugs. We don't have to visiting further child nodes to find import decl syntaxes.

This implementation does not support nested-nested-imports, though.

Sources/MockoloFramework/Models/IfMacroModel.swift
```diff
@@ -15,35 +15,88 @@
...
+            var order: Int {
+                switch self {
+                case .if:
+                    0
+                case .elseif(let order):
+                    order
+                case .else:
+                    999_999
+                }
+            }
```
fummicc1, line 52: `order` is necessary to support multiple elseIf condition.

```swift
#if FOO
import Foo
#elseif BAR
import Bar
#elseif BAZ
import Baz
#endif
```

In the example above, I added an order property to record whether BAR precedes BAZ.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -230,35 +230,46 @@ extension MemberBlockItemListSyntax {
...
         var hasInit = false
+        for (index, cl) in self.clauses.enumerated() {
+            let clauseType: IfMacroModel.Clause.ClauseType
+            if cl.condition == nil {
+                clauseType = .else
+            } else if index == 0 {
+                clauseType = .if
+            } else {
+                clauseType = .elseif(order: index)
+            }
```
sidepelican, line 244: It think using `tokenKind` is more better way

```suggestion
            let clauseType: IfMacroModel.Clause.ClauseType
            switch cl.poundKeyword.tokenKind {
            case .poundIf:
                clauseType = .if
            case .poundElse:
                clauseType = .else
            case .poundElseif:
                clauseType = .elseif(order: index)
            default:
                fatalError("unexpected tokenKind: \(cl.poundKeyword.tokenKind)")
            }
```

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -699,6 +710,8 @@ final class EntityVisitor: SyntaxVisitor {
         self.declType = declType
         super.init(viewMode: .sourceAccurate)
     }
+    
+    private static var ifConfigDeclCount: Int = 0
```
sidepelican, line 714: This is not concurrency safe. There will be times when the same value is picked and becomes randomly order.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -738,42 +751,77 @@ final class EntityVisitor: SyntaxVisitor {
...
+        imports.append(
+            .init(line: node.trimmedDescription)
+        )
         return .skipChildren
     }
 
     override func visit(_ node: IfConfigDeclSyntax) -> SyntaxVisitorContinueKind {
-        for cl in node.clauses {
-            let macroName: String
+        let blockId = Self.ifConfigDeclCount
```
sidepelican, line 761: I think you need to use file names to stabilize the order. Alternatively, you might get some useful value from Swift Syntax's `Syntax`

sidepelican (changes requested):
I am currently in the middle of a review, but I have to leave  so I’m posting this update now.

The following issues need to be addressed due to serious regressions:
- Duplicate imports
- Unstable import order

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -738,42 +751,77 @@ final class EntityVisitor: SyntaxVisitor {
...
+        Self.ifConfigDeclCount += 1
+        var previousConditions: [String] = []
+
+        for (index, cl) in node.clauses.enumerated() {
+            let compilerDirectiveKey: String
```
sidepelican, line 766: This `compilerDirectiveKey` does not seem to be just a key string, but rather a structured value. However, since it is currently a raw string, the structure it is meant to represent is unclear.  
I think it would be better to construct an `InsideDirective` directly.

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -738,42 +751,77 @@ final class EntityVisitor: SyntaxVisitor {
...
-                        if imports[key] == nil {
-                            imports[key] = []
-                        }
-                        imports[key]?.append(importItem.trimmedDescription)
-
+                        nestedImport = .init(
+                            line: importItem.trimmedDescription,
+                            compilerDirectiveKey: compilerDirectiveKey
+                        )
                     } else if let nested = el.item.as(IfConfigDeclSyntax.self) {
```
sidepelican, line 794: With this approach, you can only handle nested `#if` one time.  
I think it should be a recursive structure to work correctly.  
However, since this spec is from the beginning, maybe you can ignore it for now.

Input:

```swift
#if canImport(NewFramework)
import Z
#if canImport(NewFramework2)
import W
#if canImport(NewFramework3)
import Z
#endif
#endif
#endif
```

Current output(`import Z` has gone) :

```swift
#if canImport(NewFramework)
#if canImport(NewFramework2)
import W
#endif
import Z
#endif
```

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
sidepelican: Oh, it was clearly written here that nesting is not supported.

Sources/MockoloFramework/Templates/IfMacroTemplate.swift
```diff
@@ -17,28 +17,52 @@
...
-        
-        let template = """
-        \(1.tab)#if \(name)
-        \(rendered)
-        \(1.tab)#endif
-        """
-        return template
+        }
+
+        result += "\n\(1.tab)#endif"
```
sidepelican, line 64: nits: Maybe `enumerated` is unnecessary.

```suggestion
        result += "\(1.tab)#endif"
```

Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
```diff
@@ -738,42 +751,77 @@ final class EntityVisitor: SyntaxVisitor {
...
+        )
         return .skipChildren
     }
 
     override func visit(_ node: IfConfigDeclSyntax) -> SyntaxVisitorContinueKind {
-        for cl in node.clauses {
-            let macroName: String
+        let blockId = Self.ifConfigDeclCount
+        Self.ifConfigDeclCount += 1
+        var previousConditions: [String] = []
```
sidepelican, line 763: Only the first value is used. Is this a mistake? (Should be `String?` ?)

PR #313, opened 2025-05-23
Fix matrix destination name in release.yml

fxwx23:
It turns out this wasn’t actually an issue, since Swift 6.1 appears to be pre-installed on the [Ubuntu 22.04 image](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md).


sidepelican (approved):
Oops, Thank you 😂

PR #315, opened 2025-07-10
Add protocol-level @available annotation inheritance to generated mocks

### LINKED ISSUE #314: Generated mocks do not inherit protocol-level @available annotations
## Problem

Currently, when a protocol has `@available` annotations, the generated mock classes do not inherit these availability constraints. This leads to compilation errors when using the mocks in contexts where availability-constrained types or members are involved.

## Example

**Input protocol:**

```swift
/// @mockable
@available(iOS 18.0, *)
public protocol Foo: Sendable {
    func bar()
}
```

**Incorrect generated mock:**

```swift
public final class FooMock: Foo, @unchecked Sendable { // ❌ Missing @available
    // Mock implementation...
}
```

**Expected mock:**

```swift
@available(iOS 18.0, *)
public final class FooMock: Foo, @unchecked Sendable {
    // ✅ Correct
}
```

## Real-world Impact

When the generated mock lacks the appropriate `@available` annotation, any reference to types or members that require specific OS versions results in compile-time errors, even in purely test contexts.

## Environment

- **mockolo version:** 2.4.0
- **Xcode version:** 16.4
- **Swift version:** 6.0+

## Proposed Solution

Enhance mock generation to:

- Extract protocol-level `@available` annotations
- Apply them to the generated mock class

---

A draft PR is already available for this feature, and I would appreciate feedback based on that proposed approach:

Draft PR: 

Thank you for your time and consideration!

Keitaro0226:
## Overview

Fixes #314

This draft PR addresses an issue where `@available` annotations on protocols are not inherited by the generated mock classes. This can cause compilation errors when the protocol uses types or members that are only available in specific OS versions.

## Problem

Currently, when generating mocks for protocols with `@available` annotations, Mockolo does not apply these availability constraints to the generated mock. As a result, using the mock can trigger compile-time errors when availability-constrained types or members are involved.

### Example

```swift
/// @mockable
@available(iOS 18.0, *)
protocol Foo: Sendable {
    func bar()
}
```

Incorrect mock (missing `@available`):

```swift
public final class FooMock: Foo, @unchecked Sendable { 
    // ❌ Missing @available
}
```

Correct mock:

```swift
@available(iOS 18.0, *)
public final class FooMock: Foo, @unchecked Sendable { 
}
```

## Solution

- Extract protocol-level `@available` attributes and apply them to generated mocks.

## Benefits

- Prevents availability-related compilation errors.
- Aligns mock generation with Swift’s availability system.

---

Open to feedback!


sidepelican:
Any updates on this? When this draft PR becomes ready?

Keitaro0226:
Sorry for the long silence. I got sidetracked and lost track of this.

I'll add tests and mark this PR as ready soon. Thanks for your patience.

Tests/TestSendable/FixtureSendable.swift
```diff
@@ -208,4 +208,40 @@
         }
     }
 }
+
+@Fixture enum availableSendableProtocol {
+    /// @mockable
+    @available(iOS 18.0, *)
+    public protocol Foo: Sendable {
+        func bar() -> String
+    }
```
sidepelican, line 220: I want to enforce a compilation error when `@available` isn't applied to a mock.

```suggestion
    @available(macOS 99.0, *)
    struct Bar {}
    
    /// @mockable
    @available(macOS 99.0, *)
    protocol Foo: Sendable {
        func bar() -> Bar
    }
```

Sources/MockoloFramework/Models/ParsedEntity.swift
```diff
@@ -65,12 +65,17 @@ struct ResolvedEntity {
 
     func model() -> Model {
         let metadata = entity.metadata
+        // Combine protocol-level attributes with member-level attributes
```
sidepelican:
```suggestion
```

The meaning feels a bit off and it’s redundant, so let's remove it.

Tests/TestActor/FixtureActor.swift
```diff
@@ -100,6 +100,9 @@
             init() { }
         }
 
+        @MainActor
+        /// @mockable
```
sidepelican: It's strange for `@mockable` to come along too.
How about just inheriting `available`?
You can change `EntityNode.attributesDescription` requirement to `var attributes: [String] { get }`.

Tests/TestSendable/FixtureSendable.swift
```diff
@@ -208,4 +208,40 @@
...
+                    state.callCount += 1
+                    return state.handler
+                }
+                if let barHandler = barHandler {
+                    return barHandler()
+                }
+                return ""
+            }
+        }
+    }
```
sidepelican, line 281: I need a test case for protocol inheritance scenarios:


```swift
@available(macOS 100.0, *)
struct Bar {}

@available(macOS 90.0, *)
protocol Foo {
}

/// @mockable
@available(macOS 100.0, *)
protocol Foo2: Foo {
  var bar: Bar { get set }
}
```

sidepelican (changes requested):
Thanks for the PR. Overall it looks good, but I noticed a few minor details that I'd like you to address before we merge.

Tests/TestSendable/FixtureSendable.swift
Keitaro0226: Fixed in 2fa6c3d.
I updated the fixture to use an availability-gated return type (`Bar`) so the generated mock must inherit `@available` to compile.

Sources/MockoloFramework/Models/ParsedEntity.swift
Keitaro0226: Fixed in 3e51751.
Removed the redundant comment.

Tests/TestActor/FixtureActor.swift
Keitaro0226: Fixed in 7f3a38b.
I changed declaration attribute handling to keep syntax attributes as separate entries, so `@mockable` trivia no longer gets carried along with `@available`.

Tests/TestSendable/FixtureSendable.swift
Keitaro0226: Fixed in 32dffb1.

Keitaro0226:
Thank you for the review.
I’ve addressed the feedback and pushed the updates. I’d appreciate it if you could take another look when you have a chance.
Thank you.

sidepelican (approved):
Sorry for the late review.
It looks good. Thank you for your contribution!

Issue #307, opened 2025-05-01
Generated mock adds `@escaping` to init when dictionary has a function value


jwsinner:
Basic reproducer:

```swift
/// @mockable
protocol Tester {
  var testDictionary: [String: () -> String] { get }
}
```

Produces:
```swift
class TesterMock: Tester {
    init() { }
    init(testDictionary: @escaping [String: () -> String]) {
        self._testDictionary = testDictionary
    }

    private var _testDictionary: ([String: () -> String])! 
    var testDictionary: [String: () -> String] {
        get { return _testDictionary }
        set { _testDictionary = newValue }
    }
}
```

The `@escaping` in the init causes a compiler error due to it be associated with a dictionary rather than a function.

sidepelican:
Thanks for reporting this issue. The problem comes from Mockolo handling types as raw strings instead of structured data, which makes type operations a bit messy. Here, the type string has an arrow (->), so the dictionary is treated as a closure.

I'm working on updating SwiftType to use a structured format, which should fix this and other related issues.

sidepelican:
Fixed in https://github.com/uber/mockolo/pull/317

PR #317, opened 2025-07-16
Overhaul of `SwiftType`

sidepelican:
## Issue

`SwiftType` is essentially just a string-based type, with type-related operations performed ad hoc.

For example, determining whether a type is a closure type is done by checking for the presence of the `->` syntax.

https://github.com/uber/mockolo/blob/8c628aaa552bea925e67002dfa48e5338e2d3b26/Sources/MockoloFramework/Utils/TypeParser.swift#L124-L126

This method is extremely ambiguous and can lead to incorrect classifications—for instance, array types like `[() -> ()]` would be mistakenly treated as closures (see https://github.com/uber/mockolo/issues/307).
We want to eliminate these ad hoc, unreliable type determination methods.

## Solution

We will perform a complete overhaul of `SwiftType` to enable structured representation of Swift types.
The new `SwiftType` will systematically decompose and store various modifiers and generic parameters.
We will gain the ability to structurally examine the type's actual nature.


Sources/MockoloFramework/Utils/CoW.swift
```diff
@@ -0,0 +1,45 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+@propertyWrapper
+indirect enum CoW<Value> {
```
sidepelican, line 18: A convenient wrapper to provide CopyOnWrite semantics

Tests/TestFuncs/TestGenericFuncs/FixtureGenericFunc.swift
```diff
@@ -146,11 +146,11 @@ import MockoloFramework
             }
 
             private(set) var optionalPullCallCount = 0
-            var optionalPullHandler: (([SomeEvent], Any, Bool, ((Any?) -> ())?) -> ())?
+            var optionalPullHandler: (([SomeEvent], Any, Bool, Any?) -> ())?
             func optionalPull<T>(events: [SomeEvent], value: T, once: Bool, closure: ((T?) -> ())?) {
                 optionalPullCallCount += 1
                 if let optionalPullHandler = optionalPullHandler {
-                    optionalPullHandler(events, value, once, closure as? ((Any?) -> ()))
```
sidepelican, line 153: This cast anytime fails (except when `T` is `Any`).

Sources/MockoloFramework/Models/NominalModel.swift
```diff
@@ -47,8 +47,8 @@ final class NominalModel: Model {
          declaredInits: [MethodModel],
          entities: [(String, Model)],
          requiresSendable: Bool) {
-        self.name = name
-        self.type = SwiftType(self.name)
+        self.name = selfType.name
+        self.type = SwiftType(kind: .nominal(selfType))
```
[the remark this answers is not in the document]
fummicc1: @sidepelican 

I'm sorry I missed your reply.

> Why do you think so?

Now `SwiftType` is more structured, `SwiftType.Nominal.genericParameterTypes` is duplicated of `genericWhereConstraints` property in `NominalModel`. I though it can be not good for consistency.
However there is the following bug in current version, it might not be in this PR scope.

```swift
/// @mockable
class C<V: Sendable> {}

// actual
class CMock : C {}
```

> Even at this point, methods like SwiftType.toClosureType are already utilizing the structural information provided by enclosingType

If I understand correctly, `enclosingType` is always nominal type's typeName and there is little need to treat it as  enriched `SwiftType` because `enclosingType` is only used for rendering and checking its type with String API. In contrast, operating String is not good for developer experience. though defining `enclosingType` as `SwiftType` might be over engineering.

As the above, your PR description says structural information is more important than simplicity.

Is it difficult to ensure that type of `enclosingType` is nominal? Because `SwiftType` is common and all of type can be SwiftType, resulting the increase of combination of states.

Previously I though the following structure can be alternative.

```swift
public protocol SwiftType {
    // defined shared state and behavior
}

public struct NominalSwiftType: SwiftType {}
public struct TupleSwiftType: SwiftType {}
public struct ClosureSwiftType: SwiftType {}
public struct CompositionSwiftType: SwiftType {}
```

PR #320, opened 2025-11-02
Add support for `import` ACLs

walkingbrad:
**Background**

Swift 6.0 [adds the ability](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md) to add access level modifiers to import statements to control how those imported modules get exposed to consumers.

Mockolo does not currently respect access level modifiers on imports. The import itself does get picked up, but the ACL disappears in the generated mocks, resulting in a build failure. 

In packages that adopt the `InternalImportsByDefault` upcoming feature flag, import ACLs become required. e.g. If a `public` protocol's signature uses a type from an imported module, the import for that module must also be made `public`. So Mockolo cannot yet work in these kinds of packages.

**Solution**

1. Include the ACL when parsing imports in the related `swift-syntax` helper
2. Typify imports in `ImportsHandler` (`String` -> `struct Import`) so that we can resolve imports at a higher level
   - resolve clashing ACLs
   - sort by module name specifically

### DIFF
--- a/Sources/MockoloFramework/Version.swift
+++ b/Sources/MockoloFramework/Version.swift
@@ -3,5 +3,5 @@ public struct Version {
     public let value: String
 
     /// The current Mockolo version.
-    public static let current = Version(value: "2.4.0")
+    public static let current = Version(value: "2.5.0")
 }


Sources/MockoloFramework/Models/Import.swift
```diff
@@ -0,0 +1,178 @@
...
+            default:
+                break
+            }
+            
+            if let prefix = imp.prefix {
+                prefixByModuleName[imp.moduleName] = prefix
+            }
+            if let suffix = imp.suffix {
+                suffixByModuleName[imp.moduleName] = suffix
+            }
```
sidepelican: I don't think this logic is correct(Should write separate `#if` imports even if it's redundant), but since the original implementation was also incorrect, passing the tests should be sufficient.

PR #322, opened 2025-11-22
Generate setHandler method for actor based test-double

fummicc1:
## Motivation

When mockable protocol inherits `Actor`, its should be actor-isolated.
When we use that mock-class, setting handler method is not easy because we can't update actor-isolated properties from test site. This can be solved by adding `setHandler` methods for actor mock-class.

```swift
/// @mockable
protocol P: Actor {
    func f()
}
```

will generate: 

<table>
<thead><tr><td>Before</td>><td>After</td></tr></thead>
<tr><td>

```swift
actor PMock: P {
    init() { }


    private(set) var fCallCount = 0
    var fHandler: (() -> ())?
    func f() {
        fCallCount += 1
        if let fHandler = fHandler {
            fHandler()
        }
        
    }
}
```
</td><td>

```swift
actor PMock: P {
    init() { }


    private(set) var fCallCount = 0
    var fHandler: (() -> ())?
    func setFHandler(_ handler: (() -> ())?) {
        fHandler = handler
    }
    func f() {
        fCallCount += 1
        if let fHandler = fHandler {
            fHandler()
        }
        
    }
}
```

</td></tr>
</table>

I think `setFHandler` is helpful for users to set mock implementation.

## Implementation

If output test-double is `actor`, mockolo also renders like the following:

```swift
func setFooHandler(_ handler: ((String) async -> Result<String, Error>)?) {
    fooHandler = handler
}
```

To classify if test-double is `class` or `actor`, I added new property `mockDeclKind` to `RenderContext`.

### DIFF
--- a/Tests/TestActor/FixtureActor.swift
+++ b/Tests/TestActor/FixtureActor.swift
@@ -13,6 +13,9 @@
             }
             private(set) var fooCallCount = 0
             var fooHandler: ((String) async -> Result<String, Error>)?
+            func setFooHandler(_ handler: ((String) async -> Result<String, Error>)?) {
+                fooHandler = handler
+            }
             func foo(arg: String) async -> Result<String, Error> {
                 fooCallCount += 1
                 if let fooHandler = fooHandler {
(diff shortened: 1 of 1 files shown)


sidepelican (approved):
LGTM

PR #345, opened 2026-04-06
Bugfix/ipconfig skips protocols

AnthonyTTaylor:
Root Cause
The regression came from the IfConfigDeclSyntax visitor changes introduced by PR #328.

For non-fileMacro #if blocks, the visitor now always:

Parses the block as a ConditionalImportBlock
Returns .skipChildren
That is correct for import-only blocks, but incorrect for blocks containing declarations. In those cases, protocols, classes, and other declarations are never visited, so mockable entities are not discovered.

Fix
Only treat an #if block as a conditional import block when it contains imports exclusively.

This change adds a recursive check for whether an IfConfigDeclSyntax contains only:

import declarations
nested #if blocks that themselves contain only imports
If the block contains anything else, the visitor now returns .visitChildren so entity discovery proceeds normally.

Behavior After This Change
Import-only #if blocks continue to be parsed as conditional imports
#if blocks containing mockable protocols or other declarations are traversed
Mixed blocks fall back to child traversal
Nested #if blocks are handled recursively
Expected Outcome
Protocols annotated with ///mockable inside #if blocks once again have mocks generated, while preserving conditional import handling.

### DIFF
--- a/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
+++ b/Sources/MockoloFramework/Parsers/SwiftSyntaxExtensions.swift
@@ -786,10 +786,14 @@ final class EntityVisitor: SyntaxVisitor {
             return .visitChildren
         }
 
-        // Parse conditional import block recursively
-        let block = parseIfConfigDecl(node)
-        imports.append(.conditional(block))
-        return .skipChildren
+        if containsOnlyImports(node) {
+            // Parse conditional import block recursively
+            let block = parseIfConfigDecl(node)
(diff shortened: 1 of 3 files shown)


sidepelican:
Getting help from Copilot is good, but please first refine on your fork before proceeding.
Submit your completed PR.

AnthonyTTaylor:
> Getting help from Copilot is good, but please first refine on your fork before proceeding. Submit your completed PR.

Sorry,
Haven't used my personal account in a long time, didn't know copilot is enabled on my PR's default.

Will do.

PR #349, opened 2026-05-01
Fix nested bracket type overloading

sidepelican:
Previously, mockolo fails to generate overloads for nested types like arrays.
Consider the following code, which only produces a single `foo` function for mock class.

```swift
    /// @mockable
    protocol P {
        func foo() -> [String]
        func foo() -> [[String]]
    }
```

The issue stems from `SwiftType.displayName` continuing to use a legacy ad-hoc implementation, resulting in both `[String]` and `[[String]]` being rendered as the same `String`.
I'll modify the implementation to properly construct `displayName` in a structured manner.

### DIFF
--- a/Sources/MockoloFramework/Utils/StringExtensions.swift
+++ b/Sources/MockoloFramework/Utils/StringExtensions.swift
@@ -224,14 +224,7 @@ extension String {
     }
 }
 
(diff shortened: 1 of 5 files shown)


uhooi (approved):
LGTM!!

fummicc1 (approved):

sidepelican:
Thank you all for your reviews

PR #352, opened 2026-05-24
Fix member-level @available attributes hoisted to mock class declaration

farkasseb:
## Summary

Member-level behavioral `@available` attributes (`noasync`, `deprecated`) were concatenated and hoisted to the mock class declaration instead of staying on the individual generated members.

## Example

```swift
/// @mockable
public protocol StorageInput: Sendable {
    func put(_ value: String, for key: String) async throws
    func string(for key: String) async throws -> String

    @available(*, noasync)
    @available(*, deprecated, message: "Use put(_:for:) async throws")
    @discardableResult func set(_ value: String, for key: String) -> Result<Void, Error>

    @available(*, noasync)
    @available(*, deprecated, message: "Use string(for:) async throws")
    func string(for key: String) -> Result<String, Error>
}
```

**Before** — attributes from all deprecated members are concatenated onto the class declaration, causing a build error:
```swift
@available(*, noasync)
    @available(*, deprecated, message: "Use put(_:for:) async throws")
    @discardableResult @available(*, noasync)
    @available(*, deprecated, message: "Use string(for:) async throws")
public final class StorageInputMock: StorageInput, @unchecked Sendable {
    public var setCallCount: Int { ... }
    public func set(_ value: String, for key: String) -> Result<Void, Error> { ... }
    public func string(for key: String) -> Result<String, Error> { ... }
}
```

**After** — only the protocol-conforming methods carry their own attributes:
```swift
public final class StorageInputMock: StorageInput, @unchecked Sendable {
    public var setCallCount: Int { ... }
    public var setHandler: (...)? { ... }
    @available(*, noasync)
    @available(*, deprecated, message: "Use put(_:for:) async throws")
    public func set(_ value: String, for key: String) -> Result<Void, Error> { ... }

    public var stringForCallCount: Int { ... }
    public var stringForHandler: (...)? { ... }
    @available(*, noasync)
    @available(*, deprecated, message: "Use string(for:) async throws")
    public func string(for key: String) -> Result<String, Error> { ... }
}
```

Attributes are placed only on the protocol-conforming declaration (funcs, vars, subscripts, and inits), not on mock infrastructure (callCount, handler, state). This keeps test setup and inspection clean — configuring a handler for a deprecated method won't trigger deprecation warnings, and `noasync` won't restrict mock configuration from async tests.

One deliberate consequence: a deprecated stored property still produces a deprecation warning inside the generated convenience init (`self.data = data`) — seeding a deprecated property *is* deprecated usage, and the warning correctly points at the protocol still carrying the deprecated member.

### Note on platform availability

Existence-gating availability — `@available(iOS 16, *)`, `introduced:`, `obsoleted:`, `unavailable` — continues to hoist to the class declaration, same as before this fix, since the mock's infrastructure references the member's types unconditionally. Extended-form platform-scoped deprecations (`@available(iOS, deprecated: 12.0)`) are warning-only, so they stay on the member like their `*`-form counterparts; when a single attribute combines both (`@available(iOS, introduced: 10.0, deprecated: 12.0)`), gating wins and it hoists.

Two smaller behavior notes: non-`@available` attributes no longer leak onto the class alongside a hoisted member attribute (only `@available` is ever hoisted), and behavioral attributes on associatedtypes are deliberately dropped, since the generated typealias is referenced throughout the mock's own infrastructure.

Hoisted class-level attributes are now deduplicated per individual attribute and rendered in member source order (order-preserving `uniqued()`); previously an unordered `Set` join made the order vary between executions.

## Test plan

- 4 focused test fixtures (compiled `@Fixture` mocks, so expected output is conformance-checked by the compiler): deprecated members on the original reported issue, protocol-level + member-level `@available` coexistence, multiple attributes on one method, and platform attributes hoisting to the class
- All 155 tests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -46,6 +46,8 @@ let package = Package(
         .testTarget(
             name: "MockoloTests",
             dependencies: [
+                .product(name: "SwiftSyntax", package: "swift-syntax"),
+                .product(name: "SwiftSyntaxBuilder", package: "swift-syntax"),
                 "MockoloFramework",
                 "MockoloTestSupportMacros",
             ],
--- a/Sources/MockoloFramework/Models/Attribute.swift
+++ b/Sources/MockoloFramework/Models/Attribute.swift
@@ -14,29 +14,40 @@
 //  limitations under the License.
 //
 
-/// Represents an attribute attached to a declaration, parsed from an `AttributeList`.
-struct Attribute: Hashable {
-    enum Kind: Hashable {
-        /// Any attribute other than `@available` (e.g. `@objc`, `@MainActor`).
-        case regular
-        /// An `@available` that only affects usage diagnostics
-        /// (e.g. `@available(*, deprecated)`); kept on the generated member.
-        case behavioralAvailable
-        /// An `@available` that gates the existence of the declaration
-        /// (e.g. `@available(iOS 15.0, *)`, `introduced:`, `obsoleted:`,
-        /// platform-scoped `unavailable`); hoisted to the mock declaration
-        /// since the mock's infrastructure references the member's types
-        /// unconditionally.
-        case platformAvailable
+/// Represents an attribute attached to a declaration
+struct Attribute: Hashable, CustomStringConvertible {
+    var description: String
+
+    enum KnownKind: Hashable {
+        enum AvailableKind {
+            /// `@available(*, deprecated)`, `@available(*, noasync)`
+            case behavioral
+
+            /// `@available(iOS 26.0, *)`, `@available(iOS, introduced: 26.0)`
+            case platform
+        }
+        case available(AvailableKind)
+    }
+    var kind: KnownKind?
+
+    var isAvailable: Bool {
+        if case .available = kind {
+            return true
+        }
+        return false
     }
 
-    /// Source text of the attribute, e.g. `@available(iOS 15.0, *)`.
-    let description: String
-    let kind: Kind
+    var isBehavioralAvailable: Bool {
+        if case .available(.behavioral) = kind {
+            return true
+        }
+        return false
+    }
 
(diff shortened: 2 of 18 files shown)


sidepelican (changes requested):
It looks good, but there are several improvements.

- Create a `Attribute` struct and store the parsed information from `AttributeList` within `[Attribute]`.
    - `Model` should have this as `var attributes: [Attribute]` instead of `[String]`.
    - Add an `applyAttributeTemplate` extension to the `[Attribute]` type that functions similarly to the current `asAttributePrefix` implementation.
    - Remove `isBehavioralAvailable`, `behavioralAvailableDescriptions`, `platformAvailableString`, and `platformAvailableDescriptions`, and add computed properties to `Attribute` as needed
    - When generating objects like `VariableModel`, pass the attributes using syntax like `attributes.filter(\.isAvailable)`.
- Remove the redundant `attributes` parameter from `applyFooTemplate` and instead reference it directly from `self`.
- Avoid adding unnecessary tests. Focus only on the minimum required tests to verify functionality.
- Ensure that the generated output remains consistent across executions.

farkasseb:
Thank you so much for the review @sidepelican ! I tried to implement what you were suggesting, please let me know what you think.

Regarding the excessive testing, the trim is fine by me, but to calibrate future PRs: how do you weigh scenario coverage against fixture maintenance? My instinct was that more covered scenarios is better.

sidepelican:
Too many tests can increase maintenance costs.
When the purpose of a test is unclear, it can end up confusing developers.

sidepelican (approved):
LGTM. Since the parse logic felt a bit complex, I refactored it into a simpler version here. I believe this should be sufficient.
Thanks.

PR #353, opened 2026-06-07
Add opt-in getter call-count tracking for protocol properties

farkasseb:
## Summary

Opt-in, per-property getter call-count tracking for protocol mocks — the getter analogue of the existing setter `…SetCallCount`. Nothing is tracked unless explicitly opted in, so existing mocks are unchanged and there's no mock bloat. This is especially useful for SwiftyMocky migrations, where `Verify(mock, .once, .someProperty)` maps directly onto reading `mock.somePropertyGetCallCount`.

## Example

```swift
/// @mockable(getter: state = true)
public protocol Foo {
    var state: Int { get }
    var token: String { get }
}
```

generates:

```swift
public class FooMock: Foo {
    public private(set) var stateGetCallCount = 0
    private var _state: Int = 0
    public var state: Int {
        get { stateGetCallCount += 1; return _state }
        set { _state = newValue }   // still settable, so it can be stubbed
    }

    public var token: String = ""   // not opted in -> unchanged
}
```

Use `all = true` to opt every property in, `name = false` to opt one back out, and `--enable-getter-history` to enable it project-wide (an explicit `name = false` still wins over the flag). A `get set` property additionally bumps the existing `…SetCallCount`; async/throwing getters inject the counter into the getter body.

### Scope

Protocol mocks only (a no-op for `--mock-all` class mocks). Combine/Rx/`@Published`-intercepted publishers and `weak`/`dynamic`-modified properties are never tracked, since they can't be backed by a counting computed accessor.

## Test plan

- New `Tests/TestGetterHistory` fixtures (compiled `@Fixture` mocks, so conformance is enforced, not just string-matched) covering: specific / `all` / opt-out, get-only / get-set, defaultable / optional / non-defaultable, IUO / existential / namespaced / associatedtype types, static, global flag (including `all = false` overriding it), `--allow-set-call-count`, Combine/Rx/`@Published`/weak/dynamic exclusions, direct Rx subjects, inherited-member annotation semantics, class-mock no-op, processed-mock merge, init-collision dedup, actor, Sendable, and async/throwing getters — plus runtime tests asserting counters increment on read and stay 0 after init.
- All 179 tests pass.

---
The IUO type-rendering fix that the compiled fixtures originally surfaced was split out into #354 per review. It has since landed, and the getter-on-IUO edge-case test has been re-added here, along with rebasing on master.

🤖 Generated with [Claude Code](https://claude.ai/code)


sidepelican:
Please separate this PR into one for the bug fix and another.

farkasseb:
> Please separate this PR into one for the bug fix and another.

#354 ✅. I was debating splitting it in the first place but I only discovered the bug because of an extra test I added to this PR. So #354 should be merged first, then I can update/finalize this PR with the extra getter-on-IUO fixture.

farkasseb:
Rebased on master now that #354 has landed, and re-added the promised getter-on-IUO edge-case fixture as a separate commit: getterHistoryEdgeTypes now includes an Int! property, whose tracked getter renders a valid Int! backing store (previously this produced the Int?! that #354 fixed).

While re-testing with Fable 5, 'we' also hardened coverage with three more compiled fixtures: getter: all = false overriding --enable-getter-history (as the README promises), a tracked getter on an associatedtype property, and a test locking that inherited members follow the declaring protocol's annotation (same semantics as history:/rx:/combine:).

sidepelican:
I’m afraid I cannot accept this proposal.
We want to avoid adding more generation patterns to the codebase.
Personally, I can't think of a scenario where testing a getter's callCount is necessary, and I believe this is the first time this feature has been requested in mockolo's long history.
Since this seems to be a very specific edge case, I would recommend writing a custom mock class manually just for that specific scenario.

farkasseb:
@sidepelican Thank you so much for merging the other PRs and taking the time to review this one as well. Let me try to convince you anyway.

First, it'd be fully opt-in via annotation (or the flag), mocks that don't request it are byte-for-byte unchanged, no default-behavior change.

Second, a structural point worth naming: mockolo currently supports the setter call-count tracking (...SetCallCount / --allow-set-call-count). So this PR wouldn't add a new pattern, it'd just complete an existing one, the read-side mirror of it.

Third, I'd mostly agree with you that the "was this read?" assertions are usually redundant with the outcome but IMHO not always.

Three categories the outcome genuinely can't capture:

1. A read that must not happen: "short-circuit" paths:

```swift
final class LoginViewModel {
    private let interactor: AuthInteracting   // var authType / var isBiometricEnabled { get }

    func onViewAppeared() {
        guard interactor.authType == .biometric else { return }   // PIN -> return before touching biometrics
        if interactor.isBiometricEnabled {                        // isBiometricEnabled backed by LAContext / Keychain
            coordinator.promptForBiometrics()
        }
    }
}

---

// SwiftyMocky test for the PIN path:
Given(interactor, .authType(getter: .pin))
// mockolo: interactor.authType = .pin
sut.onViewAppeared()
Verify(interactor, .never, .isBiometricEnabled)   // proves we short-circuited
// mockolo: #expect(interactor.isBiometricEnabledGetCallCount == 0)
```

2. A read that must happen exactly once

```swift
func configure(_ rows: [Transaction]) {
    let formatter = settings.currencyFormatter            // built once, reused for every row
    cells = rows.map { Cell(amount: formatter.string(from: $0.amount)) }
}

---

// rebuilding a NumberFormatter per row is the expensive part — output is identical either way
Verify(settings, .once, .currencyFormatter)
// mockolo: #expect(settings.currencyFormatterGetCallCount == 1)
```

3. A read that must happen exactly n times

```swift
func send(_ request: Request, retries: Int) {
    for _ in 0...retries {
        var attempt = request
        attempt.token = session.accessToken          // re-read each attempt — token may have refreshed after a 401
        if transport.send(attempt).isSuccess { return }
    }
}

---

// 1 initial try + 2 retries
Verify(session, 3, .accessToken)   // re-read every attempt, not hoisted/cached once
// mockolo: #expect(session.accessTokenGetCallCount == 3)
```


Either way, thanks again to you and Uber for maintaining this project, it's saved me a ton of hand-written mocks, I really love mockolo.

sidepelican:
I understand that you prefer writing getters with side effects.
Personally, I think such code undesirable, and adding GetCallCount would actually make writing such code easier-effectively justifying its implementation.
Considering Swift's culture, I felt this feature shouldn't be included.

PR #354, opened 2026-06-07
Fix IUO type rendering so `T!` no longer generates `T?!`

farkasseb:
## Summary

Split out of #353 per review. Mockolo renders an implicitly-unwrapped optional (`T!`) as `T?!` (≈ `T??`), which doesn't satisfy the `T!` (≡ `T?`) protocol requirement — so the generated mock fails to conform.

```swift
// before:  public var x: Int?!   ❌ does not conform
// after:   public var x: Int!    ✅
```

Fixed for properties, function/subscript params & returns, and generic returns. For func/subscript handlers the synthesized closure renders `((Int?) -> Int?)?` (since `!` is illegal inside a closure type) while the signature keeps `Int!`. Inert for non-IUO types — no existing fixture output changes.

## Test plan

New `iuoVars` / `iuoFuncs` compiled `@Fixture` mocks (conformance is compiler-enforced). `swift build && swift test` green.

---
#353 carries an extra getter-on-IUO edge test that depends on this fix, so this should land first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/MockoloFramework/Utils/StringExtensions.swift
+++ b/Sources/MockoloFramework/Utils/StringExtensions.swift
@@ -119,6 +119,7 @@ extension String {
     static let arrayTypeSugarName = "[]"
     static let dictionaryTypeSugarName = "[:]"
     static let optionalTypeSugarName = "?"
+    static let optionalAutounwrappedTypeSugarName = "!"
 
     var safeName: String {
         var text = self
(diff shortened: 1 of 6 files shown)


sidepelican (approved):
Nice catch! This makes for a good bug fix.

PR #355, opened 2026-06-14
Add test case for initializer with function type in its generic type

sidepelican:
Add test cases to prevent regressions in https://github.com/uber/mockolo/issues/307


fummicc1 (approved):
LGTM!

PR #357, opened 2026-08-03
Skip directories while scanning Swift sources

### LINKED ISSUE #356: Mockolo fails when source directory contains ".swift" suffix
Given the following command and file structure, mockolo fails with an error. I think this is because mockolo attempts to interpret `Directory.swift` as a file instead of a directory because of the `.swift` suffix. We're running into this because we use the https://github.com/auth0/JWTDecode.swift package which calls itself "JWTDecode.swift".

```bash
mockolo --sourcedirs Sources
```

```
Sources
  Directory.swift
     File.swift
```

Error:

> MockoloFramework/SourceParser.swift:105: Fatal error: Retrieving contents of /Users/[...]/Sources/Directory.swift failed

Pybsama:
## Summary

- inspect enumerated resource types before passing paths to the source parser
- skip real directories and symlinks to directories while preserving symlinks to files
- add regressions for `.swift`-suffixed directories and directory/file symlink boundaries

Fixes #356

## Testing

- `swift test --filter FileScannerTests` (2 tests)
- `swift test` (158 XCTest tests and 3 Swift Testing tests)
- `swift build`
- `swift run mockolo -j 4 -s .build/issue-356-reproduction/Sources -d .build/issue-356-reproduction/Generated-swift-run.swift`

### DIFF
--- a/Sources/MockoloFramework/Utils/FileScanner.swift
+++ b/Sources/MockoloFramework/Utils/FileScanner.swift
@@ -132,12 +132,33 @@ public func scanDirs(_ paths: [String], with callBack: (String) -> Void) {
 }
 
 func scanDir(_ path: String, with callBack: (String) -> Void) {
+    let resourceKeys: Set<URLResourceKey> = [.isDirectoryKey, .isSymbolicLinkKey]
     let errorHandler = { (url: URL, error: Error) -> Bool in
         fatalError("Failed to traverse \(url) with error \(error).")
     }
-    if let enumerator = FileManager.default.enumerator(at: URL(fileURLWithPath: path, isDirectory: true), includingPropertiesForKeys: nil, options: [.skipsHiddenFiles], errorHandler: errorHandler) {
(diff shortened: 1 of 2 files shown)


Pybsama:
This is now ready for review. The upstream `master` and the PR base are still the same commit, so no rebase is needed.

Fresh verification on the exact published head with Swift 6.3.2:
- `swift test --filter FileScannerTests` — 2/2 passed
- `swift test` — 158/158 XCTest and 3/3 Swift Testing tests passed; performance test averaged 0.180 seconds
- `swift build -c release` passed
- a real `.swift`-suffixed directory generated one mock (27 LoC); a second run reported unchanged content with the same SHA-256
- `git diff --check` passed

The existing Build Actions run remains `action_required` with zero jobs, so it is waiting for maintainer approval before the fork workflow can start; it is not a test failure.

sidepelican (commented):
Looks good, but could you reduce the AI slop in the description?
Also, please avoid using double negatives in guard statements since they can be hard to read.

PR #358, opened 2026-08-12
Correct the documented default concurrency limit

KYHyeon:
## Summary

- Correct the `--concurrency-limit` help text to document the actual default of `1`.
- Keep the README command reference consistent with the CLI output.

## Background

`concurrencyLimit` is optional, and the scanner falls back to `1` when no value is provided. The existing help text incorrectly states that the default is the number of cores on the running machine.

## Testing

- `swift test`
- Verified that `.build/debug/mockolo --help` reports `(default = 1)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ OPTIONS:
   --annotation <annotation>
                           A custom annotation string used to indicate if a type should be mocked (default = @mockable). (default: @mockable)
   -j, --concurrency-limit <n>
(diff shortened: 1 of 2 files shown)


sidepelican (approved):
Indeed, based on the implementation, it appears to default to `1`.
This doesn't seem intentional, but since it has functioned this way for many years, we can consider the current implementation as the correct one.

